PR #1, opened 2026-05-07
AF 86 fork adapt boilerplate

dominykas-aai-labs:


AnaniyaT (approved):
Looks good, there are just a few linting errors left on the backend code (`make check-api`). Most of them can be resolved by `make fix-api`, and there is another that's caused by the test for `org_service.create_organization` (non-existent function currently since it's been removed). It should be ready to merge when those are fixed.

AnaniyaT (changes requested):
The tests should also pass.

PR #2, opened 2026-05-11
AF-87 setup ci pipeline

AnaniyaT:
- Add one workflow per component (`api.yml`, `ui.yml`, `openclaw-base.yml`) with `lint → test → build → publish` jobs
 - Gate `publish` job on `main` only; uses registry secrets (`REGISTRY_URL`, `REGISTRY_USERNAME`, `REGISTRY_PASSWORD`)
- Filter each workflow by path so backend changes don't trigger frontend pipelines
- Cache Docker layers (GHA), uv packages, pnpm store, and Playwright browsers
- Add per-component `VERSION` files as the source of truth for image tags
- Add placeholder `openclaw-base/Dockerfile` and `openclaw-base/smoke-test.sh` (filled in by a follow-up task)
- Add `check-ui` make target for `tsc --noEmit`
- Move `ignoredBuiltDependencies` from `pnpm-workspace.yaml` into `package.json` to fix CI install
- Pin pnpm to `@9` in `ui/Dockerfile` to match the lockfile and avoid pnpm 10 build-script errors


.github/workflows/openclaw-base.yml
```diff
@@ -0,0 +1,72 @@
...
+          context: openclaw-base
+          file: openclaw-base/Dockerfile
+          load: true
+          tags: openclaw-base:ci
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+      - name: Smoke test
+        run: |
+          docker run --rm \
+            -v "${{ github.workspace }}/openclaw-base/smoke-test.sh:/smoke-test.sh:ro" \
```
sami-bre, line 27: smoke-test failure only prevents openclaw base from being published. not api and UI ... which seems to brake the AC. pls check!

PR #28, opened 2026-06-04
AF-126 UX improvements

AnaniyaT:
- Slack tokens are now validated against the Slack API at agent create, update, and start. Invalid tokens are rejected with a specific, human-readable error (e.g. "Slack bot token is invalid", "token has been revoked") before any database writes or k8s provisioning happen. A SKIP_SLACK_TOKEN_VALIDATION config flag lets dev environments bypass this.
- In-pod healthz servers (Hermes + OpenClaw) now re-validate tokens every 5 minutes and surface failures as status: error in /healthz, so stale or revoked tokens are caught at runtime.
- Agent detail page shows an error banner with the exact failure reason when an agent is in ERROR status, and the health query now polls in that state so the message appears immediately without a reload.
- Slack config panel shows a lock icon for private channels, a "set home" button to reorder the primary Hermes channel, and a disclaimer that private channels need a manual /invite before they appear.
- Hire dialog defaults to "Set up a new bot", marks Slack as recommended, fixes button labels to match the actual Slack UI ("Install to [workspace]", "Generate Token and Scopes"), updates the provisioning  checklist to reflect what the app actually does, and freezes the progress animation on error.
- Fixed a bug where OpenClaw agents were responding to every channel message because requireMention was hardcoded to false.
- Health refetch interval reduced from 60s to 10s; health query invalidated immediately after start.
- Run frontend tests on the official playwright docker image instead of installing it in the workflow (to avoid Github runner hanging issues)


sami-bre:
Automated reviewer summary — top 5 critical issues to address before merge:

1) DB migration missing for api.domains.agents.models -> Agent.last_error
   - The PR adds Agent.last_error to the model. Add an Alembic migration that creates this nullable column before rollout to avoid runtime DB errors.

2) Slack token validation enforced by default
   - config.skip_slack_token_validation defaults to False and runtime health checks validate bot/app tokens. This will cause agents to enter ERROR in environments without tokens. Either make validation opt-in, set a safer default, or document operator upgrade steps (env vars required).

3) start.sh deletes /opt/data/.env
   - The script removes persisted .env silently (rm -f /opt/data/.env). That can delete user-managed env files. Recommend logging a warning and gating deletion with an explicit env var, or at least documenting the behavior in the upgrade notes.

4) Behavioral change: requireMention set to True for OpenClaw
   - This changes reply behavior (agents only respond when mentioned). Confirm this change is intended and document it; consider making it configurable if not universally desired.

5) Session/channel prefix rename (channel -> group) and related key changes
   - Changing _CHANNEL_PREFIXES to use group changes session keys and may break persisted session mappings. Provide a migration path or compatibility shim for existing sessions.

Suggested immediate actions:
 - Add Alembic migration for Agent.last_error.
 - Add operator-facing upgrade notes in the PR (DB migration, SKIP_SLACK_TOKEN_VALIDATION implications, .env deletion, channel/session key rename).
 - Add a log/warning or gate around deleting /opt/data/.env.
 - Confirm defaults for token validation and requireMention and document intended behavior.

I will not approve or merge — this is a review comment only. If you want, I can add targeted inline comments on the critical files.

PR #32, opened 2026-06-15
changed versioning to use appVersion from api and ui charts

dominykas-aai-labs:


helm/agentfarm-api/Chart.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: v2
 name: agentfarm-api
 description: FastAPI backend for agent-farm
 version: 0.1.0
```
AnaniyaT: Bump this

helm/agentfarm-ui/Chart.yaml
```diff
@@ -2,3 +2,4 @@ apiVersion: v2
 name: agentfarm-ui
 description: Next.js frontend for agent-farm
 version: 0.1.0
```
AnaniyaT: Bump this version too. It should be updated whenever we make changes to the helm chart.

AnaniyaT (changes requested):

PR #53, opened 2026-06-29
Af 141 connect templates and skills

andrii-zv:
Templates can now declare required skills.

**Backend**

  - Added agent_template_skill join table (migration included) linking template versions to required skills with ON DELETE CASCADE on the template side and RESTRICT on the skill side
  - TemplateRead now includes required_skills; TemplateCreate/TemplateUpdate accept required_skill_ids
  - Scrum Master predefined template seeds Jira + Confluence as required skills at startup. I decided not to add any required skills to the PR reviewer template. The reason for that is that this template requires either Bitbucket or Github. I didn't want to enforce to make both or only one skill necessary. I think it makes sense that the user will add the skill to the template based on their needs. 
  - AgentAssignedSkillRead gained a required: bool field, populated based on the agent's current template version
  - create_agent rejects with 400 if required skills are missing from skill_ids
  - update_agent rejects with 409 if required skills are in removed_skill_ids, and with 400 if re-pinning to a template whose required skills aren't covered
  - delete_skill rejects with 409 if the skill is required by any template
  - new backend integration tests were added covering all the above cases

**Update**: I have made one skill required for PR reviewer. I have made GitHub as a required one. This will be as a default required skill for PR reviewer. The logic is that if the org uses bitbucket instead of GitHub, they can just modify the template and use new version of the template which makes sense in our application.

**Frontend**
1. Template drawer
  - View mode shows a "Required skills" section with pill badges, or "None"
  - Edit mode shows all available skills as toggle pills; selection is saved in required_skill_ids on PATCH. The UI of skills part was made consistent with the UI for skills managing in the agent config.

2. Hire wizard - Skills step
  - Required skills appear first in the grid, locked-selected with a "Required by template" label
  - Hire button is disabled until all required credential fields are filled

 3. Agent skills tab
  - Required skills show a "Required" badge
  - Remove button is disabled with a "Required by template" tooltip

 4. Config drawer — Template re-pin
  - When re-pinning to a template with required skills, a "Required skills" section appears showing each skill with a "· needs X credential" hint and credential forms for any providers not already configured
  - Apply button is disabled until all required credential forms are complete
  - Required skill IDs and credentials are included in the PATCH payload

 5. Tests were added

### DIFF
--- a/api/domains/agents/service.py
+++ b/api/domains/agents/service.py
@@ -463,6 +463,16 @@ class AgentService:
                     detail="LiteLLM key generation failed; cannot create agent.",
                 ) from exc
 
+        # Validate that all template-required skills are present in the request.
+        required_ids = self.template_repository.get_required_skill_ids(template.id)
+        if required_ids:
+            missing = required_ids - set(data.skill_ids)
+            if missing:
+                raise HTTPException(
+                    status_code=status.HTTP_400_BAD_REQUEST,
+                    detail="Required template skills must be included in skill_ids",
+                )
+
         self.repository.save(agent)
 
         slack_config = None
@@ -519,16 +529,6 @@ class AgentService:
                 )
             )
             secrets.append(saved)
-
-        # Validate that all template-required skills are present in the request.
-        required_ids = self.template_repository.get_required_skill_ids(template.id)
-        if required_ids:
-            missing = required_ids - set(data.skill_ids)
-            if missing:
-                raise HTTPException(
-                    status_code=status.HTTP_400_BAD_REQUEST,
-                    detail="Required template skills must be included in skill_ids",
-                )
 
         # Resolve and validate skills before any DB writes.
         skills_to_assign = self._resolve_skills(data.skill_ids, data.secrets, org_id)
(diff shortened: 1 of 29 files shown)


api/domains/agents/service.py
```diff
@@ -507,6 +520,16 @@ def create_agent(self, data: AgentCreate, context: CurrentUserContext) -> AgentR
             )
             secrets.append(saved)
 
+        # Validate that all template-required skills are present in the request.
```
sami-bre: This check runs after the agent is created in db. Missing required skills will leave orphaned agent in db.

api/domains/agents/service.py
```diff
@@ -690,6 +730,48 @@ def update_agent(
...
+                    s.id
+                    for _, s in self.skill_repository.get_agent_skills_with_details(
+                        agent.id
+                    )
+                }
+                effective_skill_ids = (existing_skill_ids | set(data.skill_ids)) - set(
+                    data.removed_skill_ids
+                )
+                if required_ids - effective_skill_ids:
+                    raise HTTPException(
```
sami-bre: similarly here, slack/teams config is updated before skill checks. so a rejected update due to skill validation will persist slack/teams config. We don't want that.

api/domains/templates/service.py
```diff
@@ -146,7 +200,14 @@ def update_template(
             heartbeat_md=updated.get("heartbeat_md", old.heartbeat_md),
         )
         self.repository.save_template(new_template)
-        return TemplateRead.model_validate(new_template)
+        if data.required_skill_ids is None:
+            resolved_ids = list(self.repository.get_required_skill_ids(old.id))
+        else:
+            if data.required_skill_ids:
+                self._validate_skill_ids(data.required_skill_ids, org_id)
```
sami-bre, line 206: similarly here, we want to validate skills before template is saved

sami-bre (changes requested):
Here's some other issue my agent identified:

3. ui/src/features/agents/integrations.ts:33 + hire-dialog.tsx + config-drawer.tsx — INTEGRATION_PROVIDERS only covers github/jira/confluence/bitbucket, but Skill.required_providers (backend) also allows gmail/google_calendar/zoho_mail/zoho_calendar. Verified: hasIncompleteIntegration treats any unrecognized provider as permanently incomplete. If an admin marks a skill needing one of those 4 providers as template-required, the Hire and Apply-template buttons become permanently disabled with no credential field ever rendered — a full, silent dead end for that workflow, directly undercutting the "user is prompted to configure credentials" AC.

5. api/domains/skills/repository.py:100 (is_required_by_any_template) — has no "current version only" scoping. Since update_template only touches the new version's join rows, once a skill is ever required by any historical version, deletion is blocked forever, even after the current/latest template drops the requirement — with a generic error message that doesn't explain why.


6. api/api_app.py:60 vs 74 — seed_predefined_templates runs before seed_aai_cli_skills. On a fresh DB, first boot seeds scrum-master/code-reviewer with zero required skills (silently dropped via the walrus filter), contradicting the PR description's own claim. Self-heals on the next restart, so impact is limited to first-boot-of-a-fresh-environment.

andrii-zv:
**Addressed the comments**:
- Modified update_template, create_agent, update_agent so that validation would happen before any saves are done
- Added Zoho, Gmail integrations to INTEGRATION_PROVIDERS variable in the ui/src/features/agents/integrations.ts. Now, the user can provide values for them. However, validation of them is for another task in the future.
- Updated skill deletion. Now, skills can't be deleted only if they are required by the latest versions of templates. If it is required by the old versions, we delete the skill. Also, we delete rows from AgentTemplateSkill table.
- Modified api_app.py so that skills seeding would happen before template seeding.

PR #54, opened 2026-06-30
AF-130: Agent Cost Tracking

miglmigl:
Cost tracking is now live. Token spend is pulled from LiteLLM in real time and displayed per agent, with full history preserved even after agents are deleted.

**Backend**
- Added a new `api/domains/costs/` domain (`models.py`, `repository.py`, `service.py`, `routes.py`)
- Extended `LiteLLMClient.get_global_spend_report()` to aggregate per-API-key spend, daily spend, and full per-model breakdowns by parsing the `/global/spend/report` endpoint;
- Updated `LiteLLMClient.generate_key()` to include `organization_id` in metadata and format unique aliases
- Added `AgentRepository.find_all_active_for_org()` used by the cost service to distinguish live agents from deleted ones when assembling the dashboard response
- `GET /api/v1/costs/summary` returns: total org cost, per-agent cost with model breakdown and token counts, cost aggregated by model, and a daily time-series for the chart — all within a configurable date range
- Added integration tests for the new cost endpoints

**Frontend**
- New `ui/src/features/costs/` feature and `ui/src/app/dashboard/costs/` route
- **Costs dashboard** shows:
  - Total organisation spend for the selected period
  - Per-agent table with model name, input tokens, output tokens, and total cost; deleted agents appear with a "Deleted" badge and retain all their historical data
  - Cost-by-model breakdown
  - Bar chart of daily spend over time, with evenly spaced date ticks regardless of the period length; zero-spend days are shown as gaps, not omitted
- Date range filter with a calendar picker and an explicit OK button so the dashboard only refetches when the user confirms the selection, avoiding noisy intermediate requests


AnaniyaT:
- Now that we're not storing snapshots, we don't need the db migrations. 
- We need more tests to test the new features

api/infrastructure/litellm/client.py
```diff
@@ -43,13 +43,16 @@ def _headers(self, master_key: str) -> dict[str, str]:
...
-    def generate_key(self, agent_id: str) -> str:
+    def generate_key(self, agent_id: str, agent_name: str) -> str:
         master_key = self._master_key()
         url = f"{self.config.litellm_base_url}/key/generate"
         try:
             resp = httpx.post(
                 url,
-                json={"metadata": {"agent_id": agent_id}},
+                json={
+                    "key_alias": agent_name,
```
AnaniyaT: Agent name shouldn't be used as the key alias since it's not unique, an alternative can be agent_name + agent_id. Also add the org id to the metadata.

AnaniyaT (changes requested):

PR #57, opened 2026-06-30
Af 122 convo and tool call plugins

dominykas-aai-labs:


api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -0,0 +1,243 @@
...
+    from datetime import datetime, timezone
+    return datetime.now(timezone.utc).isoformat()
+
+
+def _flush():
+    with _buffer_lock:
+        if not _buffer:
+            return
+        events = list(_buffer)
+        _buffer.clear()
```
AnaniyaT, line 61: Should be cleared after success or it should be re-enqueued if the request fails. Try to make this as fault tolerable as possible.

AnaniyaT (changes requested):

PR #60, opened 2026-07-03
Af 122 convo and tool call plugins

dominykas-aai-labs:


iv3t4:
## Code Review Summary

**PR #60** — *AF-122: Push-Based Conversation & Tool Call Logging*
**Author:** @dominykas-aai-labs

Great implementation plan with well-researched hook signatures and clean domain boundaries. The dual-process approach (ingest on 8001 + main API on 8000) and cluster-only exposure are sound architectural decisions. Here are my findings:

### ⚠️ Warnings

1. **IngestService._platform_maps is never defined (service.py)**
   In `_process_messages()`, the code calls `self._platform_maps(agent)` but `_platform_maps` is not defined on `IngestService` class or module. This will raise `AttributeError` at runtime. The method exists on `ConversationService` but was not extracted/copied. This is a **blocking** bug — the entire ingest flow for messages will crash.

2. **OpenClaw plugin — message_sent hook name mismatch (index.js)**
   The PR plan documents that `message_sent` does NOT fire and recommends using `message_sending` instead. However `index.js` still registers `api.on("message_sent", ...)`. Outbound messages from OpenClaw agents will silently fail to reach the ingest endpoint. Needs `api.on("message_sending", ...)` as well or instead.

3. **OpenClaw plugin — HTTPS is not handled (index.js)**
   Uses `import http from "node:http"` for the ingest HTTP call. If `INGEST_URL` ever points to HTTPS (e.g. through a service mesh), `http.request` will fail silently. Consider using `https` module conditionally based on `url.protocol`.

4. **Hermes plugin dedup ID collisions (plugins/telemetry-push/__init__.py)**
   Message dedup IDs use `f"hermes:{session_id}:{direction}:{int(occurred_at.timestamp()*1000)}"`. Two messages from same session/direction within the same ms will collide, causing one to be silently upserted over the other. The monotonic `_counter` exists but is only used for tool call IDs, not message IDs.

5. **stop_agent now unconditionally deletes ConfigMap/Secret (service.py)**
   `self.k8s.delete_config_map(name, ns)` and `self.k8s.delete_secret(name, ns)` are now called on stop. If another process expects K8s resources to persist across stop-restart (e.g. debugging), they will be gone. The plan said the deletes were moved from `start_agent` to `stop_agent` — worth confirming this is intentional.

6. **Ingest route has no request-size limit (routes.py)**
   The `IngestBatchRequest` model has no payload size guard. A malfunctioning plugin could flush the full 500-event buffer with large tool results producing multi-MB JSON body. Consider adding middleware or a model validator to cap batch size.

### 💡 Suggestions

7. **Empty auth header behavior (routes.py:23)** — `authorization.removeprefix("Bearer ")` on empty string returns empty string which gets compared. Works but 401 would be clearer than relying on the compare.
8. **ToolCallSyncState migration (ab2b4bb88c6f)** — Verify no FK references from other tables before dropping.
9. **list_channels now DB-only (conversations/service.py)** — Running agent's channel list may be empty until the first push arrives (boot race window). Expected per design but UI should handle it.

### ✅ Looks Good

- Auth model (per-agent encrypted key + constant-time compare) is well-designed
- Ingest domain follows the existing domain pattern cleanly
- Pull-to-push migration is complete and clean — sync services fully removed
- Tests cover the key auth/process paths and builder integrations
- Kubernetes config (service port 8001 not exposed on ingress) is correct
- Buffer/flush semantics in both plugins are well-reasoned with retry + drop semantics

---
*Reviewed by Hermes Agent*

api/domains/ingest/service.py
```diff
@@ -0,0 +1,141 @@
...
+    def process(self, agent: Agent, batch: IngestBatchRequest) -> None:
+        if batch.messages:
+            self._process_messages(agent, batch)
+        if batch.tool_calls or batch.tool_results:
+            self._process_tool_calls(agent, batch)
+
+    def _process_messages(self, agent: Agent, batch: IngestBatchRequest) -> None:
+        user_map, channel_map = self._platform_maps(agent)
+
+        messages = []
```
iv3t4, line 56: In `_process_messages()` (line ~56), `self._platform_maps(agent)` is called but this method is not defined on `IngestService` — it exists on `ConversationService`. This will crash at runtime on any message batch. Needs to be copied over or extracted into a shared utility.

api/domains/agents/scripts/openclaw/plugins/telemetry-push/index.js
```diff
@@ -0,0 +1,197 @@
...
+  });
+  req.on("error", (err) => {
+    console.error(`[telemetry-push] flush error: ${err.message}`);
+    requeue();
+  });
+  req.write(body);
+  req.end();
+}
+
+function nowISO() {
```
iv3t4, line 80: The plan (and PR description) notes that `message_sent` does NOT fire in OpenClaw, and recommends using `message_sending` instead. But the code registers `api.on("message_sent", ...)` which means outbound messages will silently fail to reach the ingest endpoint. Should this be `api.on("message_sending", ...)` as a decision hook (safe for observation-only with no `{cancel: true}` return)?

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -0,0 +1,259 @@
...
+    user_id = str(getattr(source, "user_id", "") or "")
+    thread_id = str(getattr(source, "thread_id", "") or "") or None
+    text = str(getattr(event, "text", "") or "")
+    _thread_ctx_end = "[End of thread context]"
+    idx = text.find(_thread_ctx_end)
+    if idx != -1:
+        text = text[idx + len(_thread_ctx_end):].strip()
+    ts = _now_iso()
+    msg_id = f"hermes:in:{chat_id}:{int(time.time() * 1000)}:{_next_counter()}"
+    session_key = _build_session_key(chat_type, chat_id)
```
iv3t4, line 131: Message dedup IDs use millisecond timestamps: `f"hermes:{session_id}:{direction}:{int(occurred_at.timestamp()*1000)}"`. Two messages from the same session with the same direction arriving within the same ms will collide on the `(agent_id, openclaw_msg_id)` unique constraint, causing one to silently upsert over the other.

The monotonic `_counter` exists but is only used for tool call IDs. Consider incorporating it into message IDs too: `f"hermes:{session_id}:{direction}:{_next_counter()}"`.

api/domains/agents/service.py
```diff
@@ -1079,39 +1078,11 @@ def stop_agent(self, agent_id: UUID, context: CurrentUserContext) -> AgentRead:
...
-            logger.warning(
-                "%s exceeded %ss during stop of agent %s; proceeding with teardown",
-                futures[fut],
-                _STOP_SYNC_TIMEOUT_SECONDS,
-                agent_id,
-            )
-
-        self.k8s.delete_deployment(f"agent-{agent.id}", self.config.k8s_namespace)
+        name = f"agent-{agent.id}"
+        ns = self.config.k8s_namespace
```
iv3t4, line 1222: ConfigMap and Secret deletion are now unconditional on stop. If any process expects these to persist across stop-restart (debugging, manual inspection), they won't be available. The plan says moves are from start_agent to stop_agent — worth confirming this is intentional and that nothing relies on K8s resources persisting after a stop.

PR #67, opened 2026-07-09
auth loop fix

dominykas-aai-labs:
Auth loop bug fix, additionally some UI tests were failing, applied a fix there also

### DIFF
--- a/ui/src/auth/providers/auth-store.ts
+++ b/ui/src/auth/providers/auth-store.ts
@@ -14,6 +14,6 @@ export const useAuthStore = create<AuthState>((set) => ({
   authToken: null,
   isSessionExpired: false,
   setToken: (token) => set({ authToken: token, isSessionExpired: false }),
-  clearToken: () => set({ authToken: null, isSessionExpired: false }),
+  clearToken: () => set({ authToken: null }),
   expireSession: () => set({ authToken: null, isSessionExpired: true }),
 }));
--- a/helm/agentfarm-ui/Chart.yaml
+++ b/helm/agentfarm-ui/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v2
 name: agentfarm-ui
 description: Next.js frontend for agent-farm
 version: 0.5.0
-appVersion: "0.25.0"
+appVersion: "0.25.1"
--- a/ui/src/app/(auth)/layout.tsx
+++ b/ui/src/app/(auth)/layout.tsx
@@ -2,19 +2,14 @@
 
 import { type ReactNode, useEffect } from "react";
 
-import { AuthLoadingFallback } from "@/auth/components/auth-loading-fallback";
 import { useLogout } from "@/auth/hooks/use-logout";
 
 export default function AuthLayout({ children }: { children: ReactNode }) {
-  const { logout, isLoggingOut } = useLogout();
+  const { logout } = useLogout();
 
   useEffect(() => {
     void logout();
   }, [logout]);
-
-  if (isLoggingOut) {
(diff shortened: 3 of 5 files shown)


iv3t4:
## Code Review Summary

**Verdict: Comment** — Auth loop fix is solid overall. One concern about the `AuthLayout` and a minor suggestion on the `user-context-provider`.

### ⚠️ Warnings

- **ui/src/app/(auth)/layout.tsx:8** — The `useEffect` calls `void logout()` on every mount of `AuthLayout`, which happens for *every* route inside `(auth)` — including the login and register pages. For users who are *already unauthenticated* and navigating to `/login`, this triggers a needless `/api/v1/auth/logout` call (even though the `setToken(null)` in the finally block is idempotent). More importantly, when a user is *logging out from the app* and gets redirected to `/login`, the `AuthLayout` remounts and calls `logout()` *again* — this second logout call goes through `useLogout` which re-sets `isLoggingOut(true)`, but since the old `isLoggingOut` guard was removed, this is just wasted server traffic rather than a visual regression. Still, the intent of this PR was to break an auth loop, and the side-effect of extra logout calls on every `(auth)` route mount is the trade-off. Worth noting.

### 💡 Suggestions

- **ui/src/auth/providers/user-context-provider.tsx:34** — When `shouldRedirectToLogin` is true, the component shows `<AuthLoadingFallback message="Loading your account..." />` instead of redirecting to `/login`. The old code used `router.replace("/login")` for the actual redirect. Now `expireSession()` only sets `authToken: null` and `isSessionExpired: true` in the store — it doesn't trigger a navigation. Who handles the redirect? I'd expect an `<AppRouter />` or similar component that watches `isSessionExpired` to push to `/login`. If such a component exists, this is fine — but if the redirect was the old responsibility of this provider itself, then users hitting a 401 will be stuck on the loading screen indefinitely.

### ✅ Looks Good

- **Auth interceptor retry guard** (`_retried` flag) — Good fix for the infinite refresh-loop. Previously a single 401 would keep retrying refresh forever; the `_retried` flag caps it at one attempt.
- **`refreshTokens` now catches rejected shared promise** — Previously if multiple concurrent requests shared `this.refreshPromise` and it threw, the callers would get an uncaught rejection. Now each caller gets `null` back safely.
- **Decoupling `clearTokens()` from the interceptor** — Calling `onSessionExpired?.()` instead of `this.config.clearTokens()` in the interceptor's error handler is cleaner — the interceptor focuses on HTTP, the store handles state.
- **`clearToken` no longer resets `isSessionExpired`** — Previously `clearToken()` would set `isSessionExpired: false`, which meant that if the 401 error handler path called `clearToken()`, it would wipe the expired flag that `user-context-provider` relies on. Now `clearToken()` only clears the token, and the interceptor uses `onSessionExpired?.()` which properly sets the expired flag.

ui/src/auth/providers/auth-store.ts
```diff
@@ -14,6 +14,6 @@ export const useAuthStore = create<AuthState>((set) => ({
   authToken: null,
   isSessionExpired: false,
   setToken: (token) => set({ authToken: token, isSessionExpired: false }),
```
iv3t4, line 16: Since  used to reset `isSessionExpired: false`, the old path where the interceptor called `clearTokens()` on 401 would wipe the expired flag before this component even saw it. Now the interceptor calls `onSessionExpired?.()` instead, which keeps the flag true — good fix.\n\nOne concern: `clearToken()` no longer resets `isLoggingOut` either. If `useLogout` is called while a previous logout is in flight, `setIsLoggingOut(true)` runs twice but only the second finally block resets it. Not a regression this PR introduced, but worth noting.

PR #69, opened 2026-07-10
AF-173: inject aai-cli --profile mapping into auto-loaded AGENTS.md

AnaniyaT:
Agents intermittently omit --profile because the integrations mapping lived only in TOOLS.md, which neither Hermes nor OpenClaw auto-loads. Add build_integrations_policy_md and append it to agents_md at start time (both runtimes), so the per-provider --profile mapping and a no-fallback policy land in the startup system prompt.

- add build_integrations_policy_md: concise block with a policy line (aai-cli is the only path, always pass --profile, no browser/curl/HTTP fallback, don't invent URLs/tokens), one --profile <slug> line per configured provider (all 8, multi-repo variants for GitHub/Bitbucket), and a pointer to ./skills/aai-cli/*.md; empty when no integrations are configured
- append the block to agents_md in start_agent, generated from decrypted secrets only so templates cannot override or suppress it; TOOLS.md keeps its detailed block unchanged
- test the new helper and assert a configured provider's --profile line appears in the rendered AGENTS.md for both Hermes and OpenClaw


sami-bre (approved):
I can confirm multiple repository cases are handled for github/bitbucket. it's nice.

PR #74, opened 2026-07-15
Af 171 telegram support

dominykas-aai-labs:


aai-labs-agent:
## Code Review Summary

**PR #74** — AF-171: Telegram Support for Agent Farm (Hermes + OpenClaw Runtimes)
**Author:** dominykas-aai-labs
**Scope:** 46 files, +2,751 / −197 lines

This PR implements Telegram as a third messaging platform for both Hermes and OpenClaw runtimes, plus addresses 7 bugs found during live testing. The overall architecture is solid — the platform-aware approach (env var `AGENT_PLATFORM`, conditional plugin bundling, per-policy builders) is clean and consistent with the existing Slack/Teams patterns.

### 🔴 Critical

- **`agent_plans/AF_171_plan.md` committed to the repo** — This 447-line internal planning document (including bug findings, root causes, and implementation details) should not be in the repository. It leaks internal debugging notes and should be removed. Suggest using `.gitignore` for `agent_plans/` or moving to a private location.

- **Bug 4b (`_last_channel` race condition) only partially fixed** — The telemetry-push plugin (`hermes/plugins/telemetry-push/__init__.py`) only replaces the hardcoded `"slack"` string with `_PLATFORM = os.environ.get("AGENT_PLATFORM", "slack")` (line ~46). However, the plan explicitly described replacing the global `_last_channel` dict with a per-session dict keyed by `session_key` (to fix the race where concurrent sessions swap channel context). That structural fix is **absent** from the diff — the global `_last_channel` dict and its read/write pattern appear unchanged. The plan's proposed `_session_channels: dict[str, dict]` keyed by `session_key` with cleanup in `_on_session_end` was not implemented.

- **Bug 5 (`SLACK_HOME_CHANNEL`) may not be fully suppressed** — The plan explicitly states: *"The Telegram builder does not set this env var. Hermes checks it regardless of platform."* The fix described was to add `"SLACK_HOME_CHANNEL": _NO_HOME_CHANNEL` to `build_secret_hermes_telegram()`. Instead, the implementation adds `TELEGRAM_HOME_CHANNEL` and `TELEGRAM_HOME_CHANNEL_NAME`. If Hermes natively checks `SLACK_HOME_CHANNEL` as a separate env var (regardless of active platform), this could still trigger the "set home channel" prompt. Need to verify that Hermes gateway reads `TELEGRAM_HOME_CHANNEL` for Telegram agents and doesn't also check `SLACK_HOME_CHANNEL`.

### ⚠️ Warnings

- **`_telegram_maps` returns same dict for both `channel_map` and `dm_map`** (`api/domains/conversations/service.py` line ~1935: `return {}, resolved, resolved`) — Since `get_chat_display_name` is used for both user IDs and chat IDs, putting the same resolved mapping in both channel_map and dm_map means a numeric user ID could resolve as a channel name (and vice versa). This could cause incorrect display names in the UI. Consider returning a single map and letting the caller decide based on the `:dm:` vs `:group:` session prefix.

- **`dm_policy="off"` maps to `openclaw_dm_policy="allowlist"` with empty allowFrom** (`api/domains/agents/builders/openclaw.py` line ~746) — Semantically this achieves the same result (nobody is allowed), but if a human reads the OpenClaw overlay JSON, they'll see `dmPolicy: "allowlist"` when the agent was configured with `"off"`. Consider using `"allowlist"` explicitly documented as a no-op for the `off` case, or adding a comment.

- **`allowed_chat_ids` for OpenClaw uses `groups` key, not `allowedChats`** (`api/domains/agents/builders/openclaw.py` line ~781: `{"groups": {cid: {} for cid in allowed_chat_ids or []}}` ) — The plan specified `"allowedChats"` but the implementation uses `groups: {cid: {}}`. This is fine as long as OpenClaw natively reads `channels.telegram.groups` for group filtering. The REPLACE_PATHS in `init-openclaw.js` include `['channels', 'telegram', 'groups']` so the path resolution is correct — just a discrepancy from the plan.

- **`_request_json` retry loop can hit `RuntimeError("exhausted retries")`** (`api/infrastructure/telegram/client.py` line ~63) — The function catches `httpx.TransportError` and 429 inside the loop. But `resp.raise_for_status()` on a non-2xx, non-401, non-404 response will raise `httpx.HTTPStatusError` (e.g., 500, 503). That **is not** a `TransportError`, so it will propagate out of the loop. The caller's `except Exception` will catch it, but the error message will be cryptic (`"exhausted retries for /bot{token}/getMe"`). Recommend also catching `httpx.HTTPStatusError` in the inner loop, or at minimum letting `raise_for_status()` propagate cleanly.

- **`_INBOUND_TELEGRAM_RE` regex uses `(.+?)` for channel name** (`api/domains/conversations/parser.py` line ~1782) — Unlike the Slack regex which uses `(?:"([^"]+)"|(#\S+))` for channel format, the Telegram channel capture is just `(.+?)`. Telegram group names can contain special characters, emojis, etc., which should be fine with `.+?`, but worth noting this is less specific than the Slack regex.

- **`conversations-tab.tsx` empty state text still hardcoded to Slack** — The plan mentioned this as a minor fix (line 63: "Messages will appear here once the agent starts receiving Slack messages."). The diff for `conversations-tab.tsx` shows a modified file (`+28/-16` in the files list) but the plan's fix for platform-aware empty state text isn't covered. Worth checking if the diff already addresses this.

### 💡 Suggestions

- **`build_secret_hermes_telegram`** — Consider also setting a dummy `SLACK_HOME_CHANNEL` env var alongside `TELEGRAM_HOME_CHANNEL` as a defense-in-depth measure, in case future Hermes versions still check the Slack env var regardless of platform.

- **Cross-platform `validate_bot_token` pattern** — The `get_chat_display_name` function in `api/infrastructure/telegram/client.py` calls `_request_json` which has retry logic. For `getChat`, the function catches all exceptions and returns `None`. This is appropriate (degrading gracefully is better than breaking the conversation display), but the 15-second timeout per unresolved ID could accumulate if there are many IDs. With `_MAX_RETRIES = 2`, worst case is ~45 seconds per ID. Consider a batch `getChat` API call if Telegram supports it, or capping the unresolved list.

### ✅ Looks Good

- Clean separation of Telegram-specific builders (`build_hermes_config_telegram`, `build_openclaw_config_overlay_telegram`) following the existing Slack/Teams patterns
- Cross-platform field validation in `update_agent` is well-structured — using `other_platform_fields` dict instead of cascading if/else blocks
- Conditional plugin bundling in `build_hermes_config_map` (only shipping platform-relevant plugins to the pod) — clean and avoids bloat
- Scripts are conditionally loaded in `start.sh` (Hermes) via file-existence checks — safe idempotent behavior
- OpenClaw `start.sh` correctly skips `@openclaw/slack` install when `AGENT_PLATFORM != slack`
- E2E test updates to `hire-dialog.spec.ts` properly step through the updated wizard flow
- Token validation with retry logic (429 + transport error) is well-thought-out

---
*Reviewed by Hermes Agent*

agent_plans/AF_171_plan.md
```diff
@@ -0,0 +1,447 @@
+# Telegram Support for Agent Farm (Hermes + OpenClaw Runtimes)
```
aai-labs-agent: 🔴 This 447-line internal planning document (including bug findings, root causes, and implementation details) leaks internal debugging notes and should not be in the repository. Suggest adding agent_plans/ to .gitignore or moving to a private location.

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -43,9 +43,12 @@ def _buffer_push(event):
         _buffer.append(event)
 
 
+_PLATFORM = os.environ.get("AGENT_PLATFORM", "slack")
```
aai-labs-agent, line 46: 🔴 **Bug 4b not fully fixed:** The plan described replacing the global `_last_channel` dict with a per-session dict keyed by `session_key` (with `_session_channels_lock`). The diff only replaces the hardcoded `"slack"` string with `\_PLATFORM = os.environ.get("AGENT_PLATFORM", "slack")`. The structural race condition fix from Bug 4b is absent — the global `\_last_channel` dict and its read/write pattern appear unchanged.

api/domains/agents/builders/hermes.py
```diff
@@ -195,9 +279,56 @@ def build_secret_hermes_slack(
...
+            "SLACK_HOME_CHANNEL": channel_ids[0]
+            if channel_ids
+            else _NO_SLACK_HOME_CHANNEL,
             "SLACK_CHANNEL_IDS": ",".join(channel_ids),
             "SLACK_DM_ALLOWED_USERS": ",".join(allowed_dm_users),
+            "AGENT_PLATFORM": "slack",
+        },
+    )
+
+
```
aai-labs-agent, line 266: The plan says Hermes checks SLACK_HOME_CHANNEL regardless of platform. The fix should add SLACK_HOME_CHANNEL here too (the plan says: add 'SLACK_HOME_CHANNEL': _NO_HOME_CHANNEL to string_data). Instead, TELEGRAM_HOME_CHANNEL was added. Need to verify Hermes reads TELEGRAM_HOME_CHANNEL for Telegram agents.

api/infrastructure/telegram/client.py
```diff
@@ -0,0 +1,101 @@
...
+                attempt + 1,
+                _MAX_RETRIES,
+            )
+            time.sleep(wait)
+            continue
+        if resp.status_code == 401:
+            return {"ok": False, "description": "Unauthorized — invalid bot token"}
+        if resp.status_code == 404:
+            return {"ok": False, "description": "Not Found — invalid bot token format"}
+        resp.raise_for_status()
```
aai-labs-agent: resp.raise_for_status() will raise httpx.HTTPStatusError (not TransportError) for non-2xx status codes like 500/503. This won't be caught by the retry loop since it only catches TransportError — it'll propagate out and the caller's except Exception will see a misleading 'exhausted retries' message. Consider catching HTTPStatusError too or letting raise_for_status() propagate cleanly.

api/domains/agents/builders/openclaw.py
```diff
@@ -201,6 +201,97 @@ def build_openclaw_config_overlay_teams(
...
+        "models": {
+            "providers": {
+                provider: {
+                    "baseUrl": litellm_base_url,
+                    "models": [{"id": model_name, "name": model_name}],
+                }
+            }
+        },
+        "agents": {
+            "defaults": {
```
aai-labs-agent: dm_policy='off' maps to openclaw_dm_policy='allowlist' with empty allowFrom. Semantically equivalent (nobody allowed) but the overlay JSON will show dmPolicy: allowlist when the user configured off. Consider a comment explaining this mapping.

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -43,9 +43,12 @@ def _buffer_push(event):
...
+
```
aai-labs-agent, line 47: The plan's Bug 4b proposed replacing the global _last_channel dict with a per-session dict keyed by session_key:

  _session_channels: dict[str, dict] = {}

  In _on_pre_gateway_dispatch: store under _session_channels[session_key] = {...}
  In _on_post_llm_call: look up _session_channels.get(session_key_from_session_id, {})

  This diff only replaces the hardcoded 'slack' with _PLATFORM = os.environ.get('AGENT_PLATFORM', 'slack'). The structural fix of the race condition (shared global state across concurrent sessions) appears to be missing.

api/domains/agents/builders/hermes.py
```diff
@@ -195,9 +279,56 @@ def build_secret_hermes_slack(
...
+        ),
+        string_data={
+            "TELEGRAM_BOT_TOKEN": telegram_bot_token,
+            "OPENAI_API_KEY": litellm_api_key,
+            "OPENAI_BASE_URL": litellm_base_url,
+            "OPENROUTER_BASE_URL": litellm_base_url,
+            "API_SERVER_ENABLED": "true",
+            "API_SERVER_HOST": "0.0.0.0",
+            "API_SERVER_PORT": "8642",
+            "API_SERVER_KEY": api_server_key,
```
aai-labs-agent, line 295: Bug 5 (SLACK_HOME_CHANNEL): The plan says 'Hermes checks it regardless of platform.' The fix described was to add SLACK_HOME_CHANNEL to string_data to suppress the prompt. Instead, TELEGRAM_HOME_CHANNEL and TELEGRAM_HOME_CHANNEL_NAME are set here. Need to verify: does Hermes natively check TELEGRAM_HOME_CHANNEL for Telegram agents? If it still checks SLACK_HOME_CHANNEL regardless of platform, the prompt will still appear.

api/infrastructure/telegram/client.py
aai-labs-agent: resp.raise_for_status() raises httpx.HTTPStatusError (not TransportError) for non-2xx status codes like 500/503. This won't be caught by the except httpx.TransportError retry handler — it'll propagate out and the caller's except Exception will catch it with a misleading error. Consider catching HTTPStatusError in the retry loop too.

api/domains/conversations/service.py
```diff
@@ -122,6 +129,28 @@ def _platform_maps(
             logger.warning("Failed to fetch Slack maps for agent %s: %s", agent_id, e)
```
aai-labs-agent, line 129: _telegram_maps returns {}, resolved, resolved (same dict for both channel_map and dm_map). Since get_chat_display_name looks up both user IDs and chat IDs into one mapping, a Telegram user ID could resolve as a channel name or vice versa. Consider distinguishing between user vs group chat resolution — e.g., prefix user IDs with 'TL_USER:' and chat IDs with 'TL_CHAT:' in the maps.

api/domains/conversations/service.py
```diff
@@ -122,6 +129,28 @@ def _platform_maps(
...
+            self.config.agent_token_encryption_key,
+        )
+        resolved: dict[str, str] = {}
+        for chat_id in unresolved_ids:
+            raw_id = chat_id
+            if raw_id.upper().startswith("TELEGRAM:"):
+                raw_id = raw_id[len("TELEGRAM:") :]
+            name = get_chat_display_name(bot_token, raw_id)
+            if name:
+                resolved[chat_id] = name
```
aai-labs-agent, line 151: I'd separate the return for channel_map and dm_map rather than returning 'resolved' for both. get_chat_display_name resolves user IDs (first_name/username) and chat IDs (title) into the same dict, so a user ID could end up as a channel name in the UI. Consider splitting: dm_map gets user lookups, channel_map gets group title lookups.

aai-labs-agent, line 151: channel_map here gets the same resolved dict as dm_map — Telegram user ID lookups (first_name/username) are being returned as both user and channel display names. Consider separating into distinct user and chat lookups.

api/infrastructure/telegram/client.py
aai-labs-agent: resp.raise_for_status() raises httpx.HTTPStatusError (not TransportError) for non-2xx status codes like 500/503. This won't be caught by the retry loop since it only catches TransportError — it'll propagate out and the caller's except Exception will see a misleading 'exhausted retries' message. Consider catching HTTPStatusError too or letting raise_for_status() propagate cleanly.

api/infrastructure/telegram/client.py
AnaniyaT, line 1: should have some caching implemented similar to slack, there's also some code duplication. `_retry_after_seconds` and `_request_json` functions exist in the slack module as well. Maybe we should extract them to another infra module.

api/domains/agents/builders/hermes.py
```diff
@@ -105,6 +117,65 @@ def build_hermes_config(
     }
 
 
+def build_hermes_config_telegram(
```
AnaniyaT: Has a lot of duplicate code, very similar to the `build_hermes_config` function

api/domains/agents/builders/openclaw.py
```diff
@@ -201,6 +201,97 @@ def build_openclaw_config_overlay_teams(
     }
 
 
+def build_openclaw_config_overlay_telegram(
```
AnaniyaT, line 165: Same here, some duplicate code.

api/domains/agents/scripts/hermes/plugins/telegram-channel-allowlist/__init__.py
AnaniyaT, line 1: When this is set to allowlist but the chat id list is empty, it behaves the same way as 'open'

AnaniyaT (changes requested):

AnaniyaT (approved):

PR #75, opened 2026-07-16
Af 167 add audit logging

sami-bre:
AF-167: Add user-action audit logging                                                                                                                                                                                                      
                                                                                                                                                                                                                                             
  Captures who did what across the control plane (distinct from AF-5's agent conversation ingest) and surfaces it to org admins and superusers.                                                                                              
                                                                                                                                                                                                                                             
  Backend                                                                                                                                                                                                                                    
  - New audit_logs domain: append-only audit_log table (no FKs — history survives user/org deletion; actor email/name snapshotted), AuditAction string catalog (plain varchar, no PG enum → new actions need no migration), one hand-written 
  Alembic migration.                                                                                                                                                                                                                         
  - AuditLogService.record() — never raises, called post-commit; mutations record in services, reads in routes (or in the service when the target object is already in hand).                                                                
  - Captures all mutations (agents, orgs, members, templates, skills, users, auth flows) with default-deny redacted changed_fields diffs, plus significant reads (agent detail/logs/conversations/tool-calls, costs, the audit log itself)   
  with 5-min in-memory dedup so React Query refetches don't flood the log.                                                                                                                                                                   
  - Every actor→action→target triplet is a user-visible subject/object: targets carry both id and human label (agent name, org name, email); failed logins record the attempted account as target with an anonymous actor.                   
  - Read API: GET /audit-logs (paginated, filterable by action/actor/target/date, search over email+label), /audit-logs/actions, /audit-logs/export (keyset-batched CSV stream, 100k cap).                                                   
  - Scoping: org owners/admins see their own org; superusers can pass organization_id or scope=all (includes NULL-org auth/user-admin events). Members get 403.                                                                              
  - CI guard: every route must be classified in AUDITED_ROUTES or AUDIT_EXEMPT_ROUTES (with a reason) or the route-coverage test fails — new endpoints force a deliberate audit decision.                                                    
                                                                                                                                                                                                                                             
  Frontend                                                                                                                                                                                                                                   
  - New audit-logs feature: filter bar (action/search/date + CSV export), expandable table showing changed-field diffs, dashboard component with org and all-orgs variants.                                                                  
  - Org admins: audit log linked from the org management page (/dashboard/[orgId]/audit-log); superusers: global view at /dashboard/audit-logs from the admin menu.                                                                          
                                                                                                                                                                                                                                             
  Testing                                                                                                                                                                                                                                    
  - Integration: end-to-end over real HTTP + Postgres (capture shapes, org isolation, member 403, superuser scope, filters, CSV export, read dedup, visible target labels). Unit: redaction, diffing, suppression window, exception          
  swallowing, route-coverage guard.


api/migrations/versions/ba700be39a26_add_audit_log_table.py
AnaniyaT, line 1: the migration file name and the revision id don't match

api/domains/audit_logs/service.py
```diff
@@ -0,0 +1,256 @@
...
+                    changed_fields=changed_fields or None,
+                )
+            )
+        except Exception:
+            logger.exception("Failed to write audit log for action %s", action)
+
+    def _is_suppressed_read(
+        self, actor_id: UUID | None, action: str, target_id: UUID | None
+    ) -> bool:
+        key = (actor_id, action, target_id)
```
AnaniyaT: org_id should be included in the key. a user can be a member of multiple orgs.

api/domains/auth/routes.py
```diff
@@ -54,23 +59,48 @@ def login_for_access_token(
...
+        _record_failed_login()
         raise credential_exception
     if not user or not check_hash(form_data.password, user.hashed_password):
+        _record_failed_login(user.id if user else None)
         raise credential_exception
 
     token_data = TokenData(user_id=str(user.id), stamp=user.security_stamp)
     token_pair = auth_service.create_token_pair(token_data)
     _set_refresh_token_cookie(response, token_pair.refresh_token, config)
+    audit_log_service.record(
```
AnaniyaT, line 95: a super user login is recorded with is_superuser_actor = False.

AnaniyaT (changes requested):

PR #81, opened 2026-07-20
enforce bot token uniqueness

dominykas-aai-labs:


api/domains/agents/repository.py
```diff
@@ -132,9 +134,35 @@ def get_slack_config(self, agent_id: UUID) -> AgentSlackConfig | None:
...
 
+    def find_active_agent_by_bot_token_hash(
+        self, bot_token_hash: str, exclude_agent_id: UUID | None = None
+    ) -> Agent | None:
+        with Session(self.delegate.engine) as session:
+            query = (
+                select(Agent)
+                .join(AgentSlackConfig, col(AgentSlackConfig.agent_id) == col(Agent.id))
+                .where(col(AgentSlackConfig.bot_token_hash) == bot_token_hash)
+                .where(col(Agent.deleted_at).is_(None))
```
aai-labs-agent, line 345: ⚠️ This query filters by `deleted_at` but **not by `organization_id`** — every other method in this repository (`get_active`, `find_all_active_for_org`, etc.) scopes by org.

If this is a multi-tenant deployment, an agent in org B that collides with a token from org A gets an error naming org A's agent (`BotTokenConflictHTTPException` includes `conflicting.name`). That's a cross-tenant information leak.

Global uniqueness for Slack bot tokens is arguably correct (one token = one Slack app), but the error message should avoid naming the conflicting agent if it belongs to a different org. I'd pass `org_id` through `_ensure_bot_token_unique` → `find_active_agent_by_bot_token_hash` and either:
- scope the query by org (if per-org uniqueness is intended), or
- keep the query global but only include the agent name in the error when the conflicting agent is in the same org.

api/migrations/versions/a7b8c9d0e1f2_add_bot_token_hash_to_slack_config.py
```diff
@@ -0,0 +1,109 @@
...
+def upgrade() -> None:
+    op.add_column(
+        "agent_slack_config",
+        sa.Column("bot_token_hash", sa.String(64), nullable=True),
+    )
+
+    key = os.environ.get("AGENT_TOKEN_ENCRYPTION_KEY", "")
+    if not key:
+        logger.warning(
+            "AGENT_TOKEN_ENCRYPTION_KEY not set; skipping bot_token_hash backfill. "
```
aai-labs-agent: ⚠️ **Migration backfill: untested branches + silent degradation**

Three branching paths in this backfill are never exercised by integration tests (which run against an empty migrated DB):

1. **Missing key** (line 33): backfill silently skipped, rows stay unhashed, index won't catch pre-migration duplicates until a manual re-run with the key set. Consider failing the migration (raise) or adding a sentinel.
2. **Decrypt failure** (line 71): row skipped, hash stays NULL.
3. **Duplicate NULL assignment** (lines 78–83): later duplicate gets `bot_token_hash = NULL`.

Add a focused migration test that seeds pre-migration rows with duplicate and undecryptable tokens and asserts the resulting `bot_token_hash` values.

aai-labs-agent (commented):
...

api/domains/agents/service.py
```diff
@@ -491,6 +494,7 @@ def create_agent(self, data: AgentCreate, context: CurrentUserContext) -> AgentR
                     cast(str, data.slack_app_token),
                     self.config.agent_token_encryption_key,
                 ),
+                bot_token_hash=compute_bot_token_hash(cast(str, data.slack_bot_token)),
```
aai-labs-agent, line 528: ⚠️ **Orphaned agent record on race-condition IntegrityError**

`self.repository.save(agent)` at line 481 commits in its own session (the `delegate.save()` delegate opens a fresh `Session`, commits, and closes — each call is an independent transaction) **before** `self.repository.save_slack_config(slack_config)` at line 504 runs.

If the DB-level unique constraint fires `IntegrityError` at line 504 (the TOCTOU race the dual enforcement is designed to catch), the agent row from line 481 is already committed and orphaned — no Slack config, no k8s resources, but a ghost row in the `agent` table. The `BotTokenConflictHTTPException` propagates without cleanup.

Narrow window and low blast radius (no k8s resources created yet), but this PR is what makes `save_slack_config` capable of failing. Worth either reordering so the slack config is persisted first, or adding a rollback/cleanup in the `IntegrityError` handler.

aai-labs-agent (commented):
...

aai-labs-agent:
## Code Review — PR #81: Enforce bot token uniqueness

**Ticket:** AF-226 — Prevent two agents from sharing the same Slack bot identity
**Commits reviewed:** 3 (eb91ec0, 702ede9, b09ec87)
**Files changed:** 9 files, +321 −4

*Consolidated review. Supersedes the two formal reviews below — this is the authoritative summary. Inline comments on the diff reflect the final findings.*

---

### ⚠️ 1. Cross-tenant agent-name leak in conflict error

`find_active_agent_by_bot_token_hash` (repository.py:152–164) filters by `deleted_at` but **not by `organization_id`** — every other query in this repository scopes by org. The partial unique index is also table-wide, not per-org.

If this is multi-tenant, an agent in org B that collides with a token from org A gets a 409 error naming org A's agent (`BotTokenConflictHTTPException` includes `conflicting.name`). That's a cross-tenant information leak, and it's broader than the AC's org-level wording: *"Bot token must be unique across all active agents in the org."*

**Fix:** Keep the global index if desired (Slack bot tokens are globally unique credentials), but pass `org_id` through `_ensure_bot_token_unique` → `find_active_agent_by_bot_token_hash` and fall back to "another agent" in the error message when `conflicting.organization_id != org_id`.

---

### 🔴 2. docs/features/agents.md not updated (definition of done)

*(Summary-only — can't post inline on a file not changed in this diff, which is itself the problem.)*

The diff adds a new domain invariant — each active agent must have a distinct Slack bot token, create/update with a duplicate returns 409, and deletion releases the token — but `docs/features/agents.md` is unchanged. The Invariants list (line 11) and the Create section (line 38) are the authoritative doc for agent creation and Slack credential requirements.

**Fix:** Add the uniqueness invariant and its 409/delete-release semantics to the Invariants list and Create section of `docs/features/agents.md` in this diff.

---

### ⚠️ 3. Orphaned agent record on race-condition IntegrityError

In `create_agent`, `self.repository.save(agent)` (service.py:481) commits in its own session **before** `self.repository.save_slack_config(slack_config)` (line 504) runs. The `save()` delegate opens a fresh `Session`, commits, and closes — each call is an independent transaction.

If the DB-level unique constraint fires `IntegrityError` at line 504 (the TOCTOU race that the dual enforcement is specifically designed to catch), the agent row from line 481 is already committed and orphaned — no Slack config, no k8s resources, but a ghost row in the `agent` table. The `BotTokenConflictHTTPException` propagates without cleanup.

Narrow window and low blast radius (no k8s resources created yet), but this PR is what makes `save_slack_config` capable of failing in the first place. Worth either reordering so the slack config is persisted first, or adding a rollback/cleanup in the `IntegrityError` handler.

---

### ⚠️ 4. Migration backfill: untested branches + silent degradation

The backfill in `a7b8c9d0e1f2_add_bot_token_hash_to_slack_config.py:33–91` has three branching paths that integration tests never exercise — tests run against an empty migrated DB:

- **Missing `AGENT_TOKEN_ENCRYPTION_KEY`** (line 33): backfill is silently skipped. Existing rows stay unhashed; the partial unique index won't catch duplicates involving pre-migration agents until someone re-runs the migration with the key set. The `logger.warning` is easy to miss in deployment logs.
- **Decrypt failure** (line 71): row is skipped with a warning, hash stays NULL.
- **Duplicate-token NULL assignment** (lines 78–83): later duplicate gets `bot_token_hash = NULL`, logged as a data-integrity warning.

**Fix:** Add a focused migration test that seeds pre-migration rows with duplicate and undecryptable tokens, then asserts the resulting `bot_token_hash` values. For the missing-key case, consider failing the migration (raise) or adding a sentinel so operators can detect a pending backfill.

---

### ℹ️ 5. HTTP 409 vs AC's 400

The acceptance criteria says *"return a clear 400 error"* but the implementation returns 409 CONFLICT. 409 is more semantically correct for a resource conflict, and the tests are consistent with 409.

**Fix:** Either change `BotTokenConflictHTTPException` to `HTTP_400_BAD_REQUEST`, or confirm with the ticket author that 409 is an accepted refinement and update the ticket accordingly.

---

### Suppressed (verified as non-issues)

- **`BotTokenConflictHTTPException` doesn't call `super().__init__()`** — matches the established repo pattern in `api/domains/users/exceptions.py:5-9`. Not flagged.
- **Chart version not bumped** — bumping only `appVersion` (0.30.0 → 0.31.0) is correct per `operations.md` versioning rules. Not flagged.

---

### ✅ Looks Good

- SHA-256 hash column avoids storing plaintext tokens for comparison — clean approach
- Partial unique index (`WHERE bot_token_hash IS NOT NULL`) correctly allows NULL for deleted agents, enabling token reuse
- Dual enforcement (service pre-check + DB constraint) is justified — pre-check yields the friendly conflict message, index is the race-safe contract
- `IntegrityError` handler in `save_slack_config` mirrors the established pattern in `users/repository.py:134-143` and `organization_users/repository.py:35-46`
- `delete_agent` correctly nulls the hash before soft-deleting, freeing the token for reuse
- Migration backfill handles existing duplicates (keeps earliest by `created_at`, NULLs duplicates, logs data-integrity warnings)
- `bot_token_hash` is correctly kept out of `AgentRead`
- Good test coverage: create-duplicate, update-duplicate, same-token-no-conflict (self-exclusion on update), reuse-deleted-token, delete-frees-token
- `test_templates.py` and `steps/agent.py` changes are required collateral of the new constraint

---
*Reviewed by Hermes Agent + Kalkidan Betre*

PR #82, opened 2026-07-21
AF-150 Enforce permission-backed organization and agent RBAC

kalibetre:
This PR replaces direct role checks with centralized, database-backed permission evaluation. It separates organization governance from per-Agent authority, ensuring authorization is consistently enforced across Agents, memberships, templates, skills, activity, conversations, tool calls, and costs.

### Agent Access Control

Organization Owners and Admins retain implicit authority over all Agents, while Members can access only explicitly assigned Agents. New Agent Viewer, Editor, and Owner roles govern per-Agent capabilities, with APIs and UI controls for granting, updating, and revoking access. Agent creation atomically records creator provenance and grants the creator Agent Owner access.

### Cross-Cutting Permission Architecture

Organization roles now resolve through a shared permission catalogue and policy layer rather than scattered role comparisons. Repository queries apply authorization scopes directly, preventing unauthorized Agent-related activity, conversations, tool calls, and cost data from being loaded. Agent responses expose `allowed_actions` so the UI can consistently hide or disable unavailable operations.

### Technical Details

The data model introduces permissions, role grants, Agent Access Roles, and Agent Access assignments with tenant-safe constraints. Locked system roles and grants are seeded and validated at startup, while the schema supports organization-specific Agent Access Roles. Access grants use locking and idempotent conflict handling to protect against concurrent membership or Agent changes.

### Breaking Changes

- Members no longer receive organization-wide Agent visibility; they must have explicit Agent Access.
- Previously role-gated operations may now return `403`, while inaccessible Agents are generally represented as `404`.
- Deployments must apply the RBAC migration before running the updated application.
- Agent API responses now include permission-derived `allowed_actions`.

### Checklist

- [x] Add and verify the RBAC schema migration
- [x] Cover permission policies and Agent authorization with unit tests
- [x] Cover organization and Agent access flows with integration tests
- [x] Add end-to-end coverage for permission-aware UI controls
- [x] Update RBAC architecture and feature documentation


api/domains/agents/authorization.py
```diff
@@ -0,0 +1,145 @@
...
+    def require_action(
+        self,
+        context: CurrentUserContext,
+        agent_id: UUID,
+        permission: PermissionKey,
+        *,
+        detail: str = "You don't have permission to perform this action.",
+    ) -> Agent:
+        agent = self.require_visible(context, agent_id)
+        self.require_action_for_visible(context, agent, permission, detail=detail)
```
aai-labs-agent, line 87: 💡 **Suggestion — double DB query per `require_action` call**

`require_visible` queries with `AGENT_READ` scope, then `require_action_for_visible` queries again with the requested permission scope. For endpoints that call `require_action` on multiple agents (e.g., list + compute `allowed_actions`), this doubles the DB round-trips.

Consider combining into a single query that checks both visibility and the specific permission, or caching the visible agent across the two calls.

aai-labs-agent (commented):
## Code Review — AF-150 Permission-Backed Organization and Agent RBAC

**Verdict: Comment** — no blocking issues. A few suggestions below.

This is a well-structured RBAC overhaul. The separation between Organization Role permissions (catalogue-driven, fixed) and per-Agent Access Roles (database-backed, assignable) is clean, and the authorization scoping at the SQL level (`agent_scope_predicates` with correlated EXISTS subqueries) prevents unauthorized data from being loaded rather than just filtering it after the fact.

### What looks good

- **Defense in depth**: DB-level constraints (check constraints, immutability triggers, composite FKs with `ondelete=RESTRICT`) back up the service-layer authorization. The `validate_agent_access_role_scope` trigger prevents cross-tenant role injection at the DB level.
- **Atomic creator access**: `create_with_creator_access` inserts the Agent and its creator's Agent Owner access in a single transaction/session — no orphaned agents without access grants.
- **Concurrent grant handling**: `grant_access` uses `with_for_update()` row locks on both Agent and Membership before inserting, with an `IntegrityError` → re-read fallback for concurrent identical grants. Clean idempotent design.
- **404 vs 403 distinction**: Inaccessible agents are concealed as 404 (via `require_visible`), while visible-but-unauthorized actions return 403 (via `require_action_for_visible`). This prevents existence leakage.
- **Deleted-agent cost access**: The `get_agent_cost` fallback to `get_deleted_in_scope` correctly limits historical spend visibility to implicit org owners/admins only — explicit assignments don't survive soft deletion. Well-documented in the code comment.
- **Migration**: The backfill granting EDITOR to existing Members for all their org's agents is the right bridge — it preserves pre-RBAC visibility without inventing creator provenance that doesn't exist.
- **Seeder validation**: `ensure_system_catalogue` validates the exact catalogue at startup (`_validate_exact_system_catalogue`), raising `RbacSeedConflictError` if the DB has drifted. The immutability triggers prevent post-migration tampering.
- **Test coverage**: Unit tests for the policy matrix (`test_organization_role_permission_matrix_is_exact`), agent authorization (`test_agent_authorization.py`), and integration tests for RBAC flows across agents, templates, costs, and tenant resolution.

### Suggestions

1. **`require_action` issues two DB queries** (`authorization.py:86-87`) — `require_visible` queries with `AGENT_READ` scope, then `require_action_for_visible` queries again with the requested permission scope. For list-heavy endpoints this doubles the round-trips per agent. Consider a combined query or caching the visible agent.

2. **`revoke_access` doesn't use `with_for_update()`** (`repository.py:292`) — `change_access_role` locks the row before modifying, but `revoke_access` selects without locking and then deletes. A concurrent `change_access_role` could see the row as present (locked) but then fail when the delete removes it. Not a correctness bug (worst case: a 404 for the concurrent caller), but worth aligning with the locking pattern used by `grant_access` and `change_access_role`.

3. **`list_assigned_members` dict lookup could raise `KeyError`** (`access_service.py:68`) — `roles[assignments[membership.id].access_role_id]` would produce a 500 if the role isn't in the `roles` dict. The `ondelete=RESTRICT` FK prevents this in normal operation, but a `.get()` with a graceful skip would be more defensive.

### Suppressed

- **`get_agent_cost` broad `except HTTPException`** (`costs/service.py:183`): The pattern catches all `HTTPException` from `require_action` and re-raises non-404s. It's correct for the current 404/403 status codes that `require_action` emits, and the code comment documents the intent. Flagging it would be nitpicking — the pattern is self-contained and clear.
- **`_require_accepted_member` returns 409 for unverified email** (`access_service.py:203`): 409 CONFLICT for "pending member must accept invite" is slightly unusual (422 might be more conventional), but I didn't find a repo-wide convention for this status code. Not flagging without a sibling-file precedent.
- **`list_models` now requires `AGENT_CREATE`** (`service.py`): All three org roles (Owner, Admin, Member) have `AGENT_CREATE` in the catalogue, so this doesn't change behavior for any real user. Not a finding.

---
*Reviewed by Hermes Agent*

api/domains/agents/repository.py
```diff
@@ -55,52 +104,203 @@ def count_active_by_org(self, org_id: UUID) -> int:
...
+            access.access_role_id = access_role_id
+            session.add(access)
+            session.commit()
+            return access
+
+    def revoke_access(
+        self, agent_id: UUID, membership_id: UUID, organization_id: UUID
+    ) -> bool:
+        with Session(self.delegate.engine) as session:
+            access = session.exec(
```
aai-labs-agent, line 269: 💡 **Suggestion — `revoke_access` doesn't use `with_for_update()`**

`grant_access` and `change_access_role` both use `with_for_update()` to lock the Agent/Access row before modifying, but `revoke_access` selects without locking and then deletes. A concurrent `change_access_role` could see the row as present (locked) but then fail when the delete removes it.

Not a correctness bug (worst case: a 404 for the concurrent caller), but aligning with the locking pattern used by the other two methods would be more consistent.

api/domains/agents/access_service.py
```diff
@@ -0,0 +1,251 @@
...
+            for role, permissions in self.rbac_repository.list_agent_access_roles(
+                agent.organization_id
+            )
+        }
+        return [
+            self._to_assignment(
+                agent,
+                membership,
+                user,
+                roles[assignments[membership.id].access_role_id],
```
aai-labs-agent: 💡 **Suggestion — `roles[...]` dict lookup could raise `KeyError`**

`roles[assignments[membership.id].access_role_id]` would produce an unhandled 500 if the role ID isn't in the `roles` dict. The `ondelete=RESTRICT` FK on `agent_access.access_role_id` prevents this in normal operation, but a `.get()` with a graceful skip (or a `continue` in the comprehension) would be more defensive against edge cases like a race between `find_access_assignments` and `list_agent_access_roles`.

aai-labs-agent:
## Consolidated Review Summary — AF-150 Permission-Backed RBAC

*Consolidates findings from both Hermes Agent and Claude Code reviews. Supersedes the earlier formal review for the authoritative list.*

**Verdict: Comment — no blockers. Production-ready.**

---

### Findings worth tracking (by priority)

**Edge-case hardening (for future tickets — AF-216/217/218):**

1. **Ownership transfer bypasses PermissionPolicy** (`organization_users/service.py:123-129`) — `OrganizationUserService._ensure_is_owner()` uses `context.require_org_role()` directly, while `OrganizationService._ensure_is_owner()` routes through `PermissionPolicy.require()`. No impact today (outcome is identical), but when custom roles arrive (AF-216), this path won't route through the policy layer. Auditability gap.

2. **Self-revocation can leave zero explicit Agent Owners** (`access_service.py:151-162`) — `revoke_access()` has no "last explicit Agent Owner" guard. A Member with Agent Owner access can revoke their own access, leaving zero access rows. Mitigated by implicit Org Owner/Admin authority (always present today), but worth tracking for AF-216/217 when custom roles or the access management UI arrive.

**Defensive coding:**

3. **`revoke_access()` lacks `with_for_update()`** (`repository.py:288-303`) — `grant_access()` and `change_access_role()` both lock rows before modifying, but `revoke_access()` does a plain `SELECT` then `DELETE`. Benign race (worst case: concurrent caller gets 404), but inconsistent with the locking pattern of the other two methods.

4. **`list_assigned_members` raw dict lookup** (`access_service.py:68`) — `roles[assignments[membership.id].access_role_id]` would produce a 500 `KeyError` if the role isn't in the `roles` dict. The `ondelete=RESTRICT` FK prevents this in normal operation, but a `.get()` with a graceful skip would be more defensive.

5. **`require_action` issues two DB queries** (`authorization.py:86-87`) — `require_visible` queries with `AGENT_READ` scope, then `require_action_for_visible` queries again with the requested permission scope. Consider combining for list-heavy endpoints.

**Data hygiene / info disclosure (low priority):**

6. **Migration backfill includes soft-deleted agents** (`migration:285-303`) — The backfill INSERT doesn't filter `agent.deleted_at IS NULL`, creating inert stale `AgentAccess` rows for deleted agents. Harmless — scope predicates exclude them from all queries.

7. **Error message leaks user/org UUIDs** (`auth/utils.py:105-107`) — `ForbiddenException(detail=f"User {user.id} does not have access to organization {organization_id}")` confirms org existence to authenticated users probing UUIDs. Minor info disclosure at auth boundaries.

---

### Confirmed strengths

- **Defense in depth**: Application-level checks + 4 DB triggers + composite FKs with `ondelete=RESTRICT`
- **SQL-level authorization**: `agent_scope_predicates` with EXISTS subqueries — unauthorized data never loaded, not just filtered post-fetch
- **404/403 boundary**: Inaccessible agents concealed as 404 (no existence leakage), unauthorized actions as 403
- **Atomic creator access**: `create_with_creator_access` — single transaction for agent + access grant
- **Concurrent grant handling**: `FOR UPDATE` locks on Agent + Membership + `IntegrityError` fallback → idempotent
- **Permission matrix verified correct**: All 21 permissions across 3 org roles + 3 agent access roles match spec
- **Seeder validation**: Startup check raises `RbacSeedConflictError` if DB catalogue has drifted
- **Migration safety**: `ON CONFLICT DO NOTHING`, complete downgrade path
- **Test coverage**: 835 API tests, 162 Playwright tests, RBAC-specific unit + integration suites
- **Admin escalation prevention**: Owner role can't be set via `add_member`/`change_role`, admin promotion/demotion is owner-only, ownership transfer is owner-only

### Monitor post-merge

1. **EXISTS subquery performance for Members** — every agent query for a plain Member includes an EXISTS subquery joining `agent_access` + `agent_access_role_permissions`. Indexes should cover it, but monitor query plans as Member population grows.
2. **Audit trail gap** — access changes (grant, revoke, role change) have no audit logging. Deferred to AF-218.
3. **Migration speed on large datasets** — backfill is `O(members × agents)` per org. Safe to retry via `ON CONFLICT DO NOTHING`, but could be slow on large orgs.

---
*Consolidated by Hermes Agent — findings from Hermes Agent + Claude Code reviews*

api/domains/agents/authorization.py
kalibetre: this only for actions, list queries are not affected.

api/domains/agents/repository.py
kalibetre: with_for_update added

api/domains/agents/access_service.py
kalibetre: addressed

PR #85, opened 2026-07-22
AF 160 shared credentials

dominykas-aai-labs:
Shared credentials are created by OWNER/ADMIN.
The page can be found in Settings -> Shared credentials.
All members in the org can attach the shared credentials.


kalibetre (commented):
### Summary

 │ Worst Standards issue: Shared-vs-manual provider overlap check raises HTTP 422 (schema validation) instead of 400 for a business precondition failure — violates docs/guidelines/code.md HTTP semantics.
 │ Worst Spec issue: Zoho Mail is listed as a supported shared-credential provider but has no validator, so validating a Zoho Mail shared credential returns a generic HTTP 400 — the "supported provider" invariant in docs/features/integrations.md is only partially honoured.

 ### Standards

 ### Standards Findings (6)

 Finding 1 — Hard
 - File:Line: ui/src/features/agents/components/agent-skills-tab.tsx:22-24[-25], ui/src/features/agents/components/hire-dialog-steps.tsx:26-28[-29]
 - Category: docs/guidelines/webapp.md → Imports and naming → "Order imports as external packages, @/* aliases, then relative imports"
 - What: Three @/*-aliased imports (SharedCredentialPicker, SHARED_CREDENTIAL_PROVIDER_LABELS, SharedCredentialBrief) are placed after the relative ../integrations import and before further relative imports, breaking the required external → @/* → relative ordering in both files.
 - Fix: Move the three @/* imports up into the @/* import block (before the first relative import) in both files; in hire-dialog-steps.tsx also move the trailing @/components/ui/tooltip import (line 33) into the @/* block.

 Finding 2 — Hard
 - File:Line: api/domains/agents/service.py:581
 - Category: docs/guidelines/code.md → HTTP semantics → "400 for business precondition failures" / "422 for schema-driven FastAPI/Pydantic validation"
 - What: The shared-vs-manual provider overlap check in the create path raises HTTP_422_UNPROCESSABLE_ENTITY for a business precondition ("Provider already has a manual credential") that requires DB access to evaluate, not schema-driven Pydantic validation.
 - Fix: Change the status code to status.HTTP_400_BAD_REQUEST to match the documented mapping for business precondition failures.

 Finding 3 — Smell
 - File:Line: api/domains/shared_credentials/service.py:36-41, api/domains/agents/service.py:127-131
 - Category: Duplicated Code
 - What: The _VALIDATORS dict mapping SecretProvider to validator functions is defined identically in both shared_credentials/service.py and agents/service.py, and will drift when a provider validator is added or changed.
 - Fix: Extract the _VALIDATORS mapping into api/infrastructure/integration_validators/__init__.py (or a shared module) and import it in both services.

 Finding 4 — Smell
 - File:Line: api/domains/shared_credentials/service.py:201-211, api/domains/agents/service.py:1581-1591
 - Category: Duplicated Code
 - What: The validation-result formatting block (computing validation_status from result.valid/result.missing_scopes and assembling the return dict) is copy-pasted verbatim between the two service validate_* methods.
 - Fix: Extract a shared helper (e.g. format_validation_result(result) -> dict) in the integration validators package and call it from both methods.

 Finding 5 — Smell
 - File:Line: ui/src/features/agents/components/agent-skills-tab.tsx:127-155, ui/src/features/agents/components/hire-dialog-steps.tsx:1503-1532, ui/src/features/agents/components/hire-dialog-steps.tsx:1703-1732
 - Category: Duplicated Code
 - What: The switchToShared/switchToManual/handlePickShared handler triad and the shared/manual toggle-button JSX with identical inline styles are duplicated across AgentSkillsTab, SkillsStep, and IntegrationsStep.
 - Fix: Extract a shared SharedManualToggle component (and optionally a useSharedManualSwitch hook) under ui/src/features/shared-credentials/components/ and reuse it in all three locations.

 Finding 6 — Smell
 - File:Line: api/domains/agents/models.py:436-441
 - Category: Speculative Generality
 - What: validate_no_shared_and_manual_overlap is a model_validator that does nothing — its body is return self with a comment stating it is a no-op placeholder — adding a dead validator that future readers may mistake for active logic.
 - Fix: Remove the no-op validator entirely; the overlap check already lives in the service where DB access is available, so the placeholder adds no value.

 ### Spec

 ### Spec Findings (2)

 Finding 1 — Suspicious
 - File:Line: api/domains/agents/service.py:878-905
 - Category: Suspicious Implementation
 - What: When an agent update sends both a manual secret and a shared credential for the same provider in one request, the shared credential silently overwrites the manual secret with no error, whereas the create path explicitly rejects this overlap with HTTP 422.
 - Spec: "An agent gets either a shared credential or a per-agent secret for a given provider, not both."
 - Fix: Add a provider-overlap check in the update shared-credential block (mirroring the create-path check at service.py:582-587) that raises HTTP 422 when the upserted manual secrets and shared credentials share a provider.

 Finding 2 — Suspicious
 - File:Line: api/domains/shared_credentials/service.py:31-36
 - Category: Suspicious Implementation
 - What: The _VALIDATORS map omits Zoho Mail, so validate_shared_credential returns HTTP 400 "No validator available for zoho_mail" even though Zoho Mail is listed in SHARED_CREDENTIAL_ALLOWED_PROVIDERS as a supported shared-credential provider.
 - Spec: "Only manual-entry providers are supported for shared credentials (v1): GitHub, Jira, Confluence, Bitbucket, Zoho Mail."
 - Fix: Either add a Zoho Mail validator to _VALIDATORS or, if live validation is not available for Zoho Mail yet, return a clear "validation not supported for this provider" message instead of a generic 400 that implies the provider itself is unsupported.

api/migrations/versions/c4d5e6f7a8b9_add_shared_credential_table.py
```diff
@@ -0,0 +1,80 @@
...
+from sqlalchemy.dialects import postgresql
+
+revision: str = "c4d5e6f7a8b9"
+down_revision: Union[str, None] = "a7b8c9d0e1f2"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # 1. Create shared_credential table
```
kalibetre, line 22: **Standards — Hard** (`docs/guidelines/code.md` — schema changes MUST include a valid Alembic migration)

The new migration sets `down_revision = "a7b8c9d0e1f2"`, which already has two other children (`e4f5a6b7c8d9`, `e7a4b9c2d5f1`) merged via `ad8b4a278dd1`. Nothing references `c4d5e6f7a8b9` as a `down_revision`, so the migration graph now has **two heads** (`ad8b4a278dd1` and `c4d5e6f7a8b9`) and `alembic upgrade head` will fail with "Multiple heads are present."

**Fix**: Add a merge migration joining `c4d5e6f7a8b9` with `ad8b4a278dd1`

api/tests/integration/test_shared_credentials.py
```diff
@@ -0,0 +1,441 @@
...
+            assert_that(response.status_code, equal_to(status.HTTP_409_CONFLICT))
+
+
+def test_create_shared_credential_without_auth_returns_401():
+    with given(_GIVEN) as context:
+        client: TestClient = context.client
+
+        with when("I create a shared credential without auth"):
+            response = client.post(
+                _BASE,
```
kalibetre, line 170: **Standards — Hard** (`docs/guidelines/testing.md` — API behavior changes MUST cover auth & role/tenant failures)

The suite covers happy path, 401 (no auth), validation, 404, and 409, but has **no test asserting a non-admin member receives 403** on create/update/delete, and **no cross-organization tenant-isolation test** (user in org A cannot read/list org B's credentials).

**Fix**: Add integration tests for member-role 403 on the mutating endpoints and cross-org 404/empty-list on get/list/briefs.

api/domains/shared_credentials/repository.py
```diff
@@ -0,0 +1,127 @@
...
+            if "uq_shared_credential_org_name" in str(e).lower():
+                from fastapi import HTTPException, status
+
+                raise HTTPException(
+                    status_code=status.HTTP_409_CONFLICT,
+                    detail=f"A shared credential named '{credential.name}' already exists in this organization",
+                )
+            raise
+        return credential
+
```
kalibetre, line 33: **Standards — Smell** (Divergent Change / layering — `docs/guidelines/code.md`: "Domain error translation SHOULD happen in services, not routes")

The repository raises a raw `fastapi.HTTPException` for the unique-constraint conflict. The established pattern (`agents/repository.py` → `BotTokenConflictHTTPException`, `users/repository.py` → `EmailTakenHTTPException`) uses named domain exception classes.

**Fix**: Introduce a `SharedCredentialNameConflictHTTPException` (or raise a domain exception translated in the service) to match the existing convention.

api/domains/agents/service.py
```diff
@@ -882,6 +921,38 @@ def update_agent(self, agent_id: UUID, data: AgentUpdate, context: CurrentUserCo
                         )
                     )
 
+        # Shared credential attachments
+        if "shared_credentials" in updated:
+            upserted_manual_providers = {item.provider for item, _ in upserts} if "secrets" in updated else set()
```
kalibetre: **Standards — Smell** (Edge-case inconsistency vs. the create path at `:606-:614`)

On agent update, the shared-credential attach conflict check (`upserted_manual_providers`) only considers manual secrets upserted in the *same request*. An **existing** manual secret for the same provider is silently overwritten (`existing.content = None; existing.shared_credential_id = shared_cred.id` at `:941`) rather than rejected with 400 as the create path does.

**Fix**: Extend the conflict set to include all current manual providers for the agent (`{s.provider for s in current_secrets if not s.shared_credential_id}`) so attaching a shared credential for a provider that already has a manual secret raises 400 instead of silently replacing it.

api/domains/agents/service.py
```diff
@@ -590,8 +602,34 @@ def create_agent(self, data: AgentCreate, context: CurrentUserContext) -> AgentR
             )
             secrets.append(saved)
 
+        # Shared credentials: look up each, verify org ownership, create link rows
+        for attach in data.shared_credentials:
```
kalibetre, line 634: **Spec — Missing** (spec invariant: "Duplicate providers in create/update payloads are rejected.")

Duplicate providers within the `shared_credentials` attach list are **not rejected** — if two shared credentials for the same provider appear in a single create/update payload, the second silently overwrites the first via `get_secret` + `save_secret` instead of raising 400. Applies to both the create path here and the update path at `:927`.

**Fix**: Add a service-level check that rejects duplicate providers among `data.shared_credentials` (and across the combined secrets + shared_credentials set), raising HTTP 400 before any DB write.

api/domains/shared_credentials/repository.py
```diff
@@ -0,0 +1,127 @@
...
+            return []
+        with Session(self.delegate.engine) as session:
+            query = (
+                select(SharedCredential)
+                .where(col(SharedCredential.id).in_(ids))
+                .where(col(SharedCredential.organization_id) == org_id)
+            )
+            return list(session.exec(query).all())
+
+    def get_many_by_ids(self, ids: list[UUID]) -> list[SharedCredential]:
```
kalibetre: **Standards — Smell** (`docs/features/rbac/IMPLEMENTATION-BRIEF.md` — visibility enforced in the repository query, defense-in-depth)

`get_many_by_ids` fetches shared credentials by ID **without filtering by `organization_id`**. All user-facing write paths (attach) correctly use `get_by_id_and_org`, so this is not directly exploitable, but the agent-read and agent-start decryption paths (`agents/service.py:329`, `:364`, `:1187`, `:1642`) trust the `agent_secret.shared_credential_id` FK without re-checking org scope — a latent cross-org credential-read risk if a bad reference is ever persisted.

**Fix**: Add an `org_id` parameter to `get_many_by_ids` (or use the existing `get_by_ids_and_org`) at the agent-service call sites so org scope is always enforced on read.

api/domains/shared_credentials/routes.py
```diff
@@ -0,0 +1,100 @@
...
+    service: Annotated[SharedCredentialService, Injected(SharedCredentialService)],
+):
+    service.delete_shared_credential(credential_id, context)
+    return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+
+@shared_credentials_router.post("/{credential_id}/validate")
+def validate_shared_credential(
+    credential_id: UUID,
+    context: Annotated[CurrentUserContext, Depends(get_current_user())],
```
kalibetre: **RBAC — Hard** (`docs/features/rbac/IMPLEMENTATION-BRIEF.md` — Shared Organization resources; `docs/guidelines/code.md` — Authorization enforcement)

The `validate` endpoint uses `get_current_user()` with **no organization-role restriction**, and the service calls only `_org_id(context)` (any member), not `_require_manager`. So **any org member** can trigger `validate_shared_credential`, which **decrypts the credential content** (`decrypt_content` at `service.py:176`) and makes a **live outbound API call** to the external provider using the org's secret.

This is a sensitive operation, not a read: it exercises admin-managed secrets externally and can probe validity / trigger outbound calls from the server identity. The RBAC brief's Shared Organization Resources principle restricts management of shared definitions to Owner/Admin; live validation of a decrypted secret is closer to "manage" than "read".

**Spec**: `docs/features/integrations.md:21` — "only admins (owner/admin roles) can create, update, or delete them" (validate is unspecified, but decrypting+exercising a secret should default to admin-only).

**Fix**: Gate the route with `Depends(get_current_user(organization_roles=IMPLICIT_AGENT_OWNER_ROLES))` and call `self._require_manager(context)` at the top of `validate_shared_credential`, matching create/update/delete.

kalibetre:
Staging has replaced `is_superuser` with a **platform admin** model. This PR is behind and will conflict in `api/domains/agents/service.py`, `api_app.py`, test steps, and UI hooks. After updating:

- Replace `is_superuser` → `is_platform_admin`, `require_superuser` → `require_platform_admin`, `check_superuser` → `check_platform_admin` in any code/tests this branch touches.
- Note the semantic change: `has_org_role` **no longer short-circuits** for platform admins — org-scoped authority now requires a real Membership. So `_require_manager` will correctly 403 a platform admin without an Owner/Admin Membership in the org. That's the right behavior for an org-scoped resource.
- Add a test: **platform admin without org membership gets 403** on `POST/PATCH/DELETE /shared-credentials` (mirror `test_membershipless_platform_admin_cannot_manage_agent_share_in_org_url`).
- Re-check the migration `down_revision` against staging's current Alembic head — staging may have moved the chain.

The inline review comments (validate-endpoint admin gate, duplicate-provider rejection, missing 403/cross-org tests, `get_many_by_ids` org-scope) still apply after the update.

api/domains/agents/service.py
kalibetre, line 152: "shared_credentials" should be added to this list right? so that so attaching/detaching a shared credential requires `AGENT_SECRET_MANAGE` like every other credential mutation.

api/domains/shared_credentials/service.py
```diff
@@ -0,0 +1,191 @@
...
+        cred_filter: SharedCredentialFilter,
+        pagination: Pagination,
+        context: CurrentUserContext,
+    ) -> PaginatedItems[SharedCredentialRead]:
+        org_id = self._org_id(context)
+        credentials, total = self.repository.find_all_for_org(org_id, cred_filter, pagination)
+        reads = []
+        for cred in credentials:
+            agent_count = self.repository.count_agent_references(cred.id)
+            reads.append(self._to_read(cred, agent_count=agent_count))
```
kalibetre: this will call the count agent ref (a db call) once per credential. basically an N+1 issue here. One option is to compute agent-reference counts for the whole page in one grouped query and join the counts onto the paginated rows in the repository.

kalibetre:
Few things to clean up as well

1. The same ~60-line block (repo-list / secret / radio / plain-text field switch, plus the `SharedManualToggle` wrapper and `{!useShared && (...)}` gate) appears **three times**:

1. `ui/src/features/agents/components/agent-skills-tab.tsx:319-437`
2. `ui/src/features/agents/components/hire-dialog-steps.tsx:1616-1750` (`SkillsStep`)
3. `ui/src/features/agents/components/hire-dialog-steps.tsx:1807-1949` (`IntegrationsStep`)

2. `useSharedManualSwitch` centralizes switch-to-shared/switch-to-manual/pick-shared logic but is only consumed by `agent-skills-tab.tsx`; `SkillsStep` and `IntegrationsStep` each re-implement the identical transitions inline instead of calling the hook. Reuse `useSharedManualSwitch` in `SkillsStep` and `IntegrationsStep`, or delete the hook if inlining is intentiona

PR #86, opened 2026-07-22
AF-231 Add Agent General Access for Dynamic Role-Based Agent Visibility

kalibetre:
This PR introduces **Agent General Access**, a new Agent-level setting that grants a chosen Agent Access Role to all current and future accepted Organization Members. It operates as an additive authorization source alongside explicit per-Member Agent Access, enabling owners to grant baseline permissions without materializing individual assignments.

## Agent General Access Backend

- Added `general_access_role_id` nullable foreign key column on the `agent` table (new Agents and migrated ones default to `NULL`, meaning **Restricted**).
- New access-management endpoints under `/agents/{agent_id}/general-access`:
  - `GET` — read the current general access role (or `null`).
  - `PUT` — set a role (must belong to the same organization or be a system role and grant `agent.read`).
  - `DELETE` — remove general access (reverts to Restricted).
- Only roles carrying `agent.read` are accepted (enforced in service layer); custom roles from other organizations are rejected by a database trigger.
- The setting is mutable only by users with `agent.access.manage` permission.

## Union-Based Authorization

- Agent visibility and effective permission resolution now combine explicit Agent Access with Agent General Access using an SQL `OR` clause in repository queries.
- The union is additive – neither source can subtract permissions from the other. If a member has both a direct Editor role and a general Viewer role, they retain Editor-level capabilities. Removing either source leaves the other intact.
- Pending or unverified Memberships receive **no** permissions from Agent General Access (controlled via a new `include_general_access` flag in `AuthorizationScope`, active only when `email_verified_at` is set).
- Organization Owners/Admins continue to have implicit Agent Owner authority unaffected by this change.

## Database Constraints & Migration

- `ON DELETE RESTRICT` on the foreign key protects referenced roles from deletion while referenced.
- A PL/pgSQL trigger enforces that custom (non-system) roles must belong to the same organization as the agent.
- The migration is additive and backward-compatible: existing agents remain Restricted, no data backfill needed.
- Downgrade cleanly removes the column, trigger, and index.

## Technical Details

- The `authorization_scope` model gained `include_general_access: bool` (default `True`), used by `agent_scope_predicates` and `get_agent_batch_permissions` to toggle the OR join on general access.
- Batch permission queries now union results from the direct `AgentAccess` table and the `Agent.general_access_role_id` join path; performance impact is bounded by a single extra query per batch.
- The `AgentGeneralAccessUpdate` schema accepts a single `access_role_id` field; the service validates role existence, org scope, and the `agent.read` permission before persisting.


aai-labs-agent:
## Code Review — AF-231: Agent General Access

**Verdict: Comment** — no blocking issues. Implementation is well-structured with solid test coverage and documentation. One documentation discrepancy to address.

---

### Summary

This PR introduces Agent General Access — an Agent-level setting that grants one Agent Access Role to all current and future accepted Organization Members. The implementation replaces the old granular access API (`POST`/`PATCH`/`DELETE /access`) with a consolidated share snapshot API (`GET`/`PUT /share`) that atomically replaces both general access and explicit assignments in a single transaction.

The core design is sound: a nullable FK on the `agent` table (`general_access_role_id`) with `ON DELETE RESTRICT`, a PL/pgSQL trigger enforcing same-org scope for custom roles, and an additive `OR`-based authorization union in repository queries. The `include_general_access` flag on `AuthorizationScope` correctly gates general access for pending/unverified members.

### ✅ Looks Good

- **Migration**: Clean additive migration — nullable column, FK with `RESTRICT`, index, and a trigger that validates role existence and same-org scope. Downgrade is symmetric.
- **Atomic replace pattern**: `replace_access_settings` in the repository uses `with_for_update()` row locks on agent, roles, and memberships, then diffs and applies changes in a single transaction. This eliminates the race conditions that the old per-operation `grant_access`/`change_access_role`/`revoke_access` methods had to handle individually.
- **Union-based authorization**: `agent_scope_predicates` correctly builds `or_(direct_access, general_access)` when `include_general_access` is True, and falls back to direct-only when False. `find_agent_permissions` runs two queries and merges results into a set — the union is genuinely additive.
- **Defense-in-depth**: Cross-org role prevention is enforced at three levels — service validation (`_require_access_role` scopes by org), DB trigger (fires on INSERT/UPDATE), and FK constraint (role existence). `agent.read` requirement is enforced in the service layer (`_require_general_access_role`).
- **Pending/unverified member gating**: `include_general_access=context.user.email_verified_at is not None` in `authorization.py` ensures pending members get no general access, satisfying AC #3.
- **Test coverage**: 16 new tests in `test_agent_general_access.py` + 5 schema/migration tests in `test_rbac_schema.py` + updated `test_agent_rbac.py` suite. All 16 AC items are covered (see cross-check below).
- **Documentation**: ADR with 4 considered alternatives, updated feature docs, changelog, implementation brief, and glossary term — all accurate and consistent with the implementation.

### 💡 Notes

**1. PR description doesn't match the implemented API**

The PR body describes separate endpoints under `/agents/{agent_id}/general-access` with `GET`/`PUT`/`DELETE`, and references a model called `AgentGeneralAccessUpdate`. The actual implementation uses:
- `GET /agents/{agent_id}/share` → returns `AgentAccessSettingsRead` (general access + assignments)
- `PUT /agents/{agent_id}/share` → accepts `AgentAccessSettingsUpdate` (replaces both atomically; `general_access_role_id: null` removes general access)

The repo docs (`agents.md`, `IMPLEMENTATION-BRIEF.md`) correctly describe the actual implementation, so this is just the PR body that's stale. Worth updating to avoid confusing the AF-217 UI team or other reviewers.

**2. No explicit test for Editor/Owner general access roles**

AC #8 mentions "General access supports locked Viewer, Editor, and Owner roles." Tests cover Viewer (locked) and a custom role with `agent.read`, but don't explicitly set general access to Editor or Owner and verify the enhanced permissions (e.g., that Editor general access grants `agent.update`). The permission resolution code is generic — it checks if the role's permissions include the required one — so the mechanism is already exercised. This is a minor coverage gap, not a correctness risk.

---

### AC Cross-Check

| # | Acceptance Criterion | Status | Evidence |
|---|---|---|---|
| 1 | Each Agent has General access set to Restricted or All Org Members with one role | ✅ | `general_access_role_id` nullable FK; `NULL` = Restricted |
| 2 | New Agents default to Restricted; migration leaves existing Agents Restricted | ✅ | `default=None` on model; `test_upgrade_leaves_existing_agents_restricted` |
| 3 | General access applies to accepted Members; pending/removed receive nothing | ✅ | `include_general_access=email_verified_at is not None`; `test_pending_members_do_not_receive_general_access`, `test_removed_membership_receives_no_general_access` |
| 4 | General + direct are additive; neither reduces the other | ✅ | `or_()` in predicates; set union in `find_agent_permissions`; `test_direct_and_general_access_permissions_are_additive` |
| 5 | Removing one source leaves the other intact | ✅ | `test_direct_and_general_access_permissions_are_additive`, `test_removing_general_access_leaves_direct_access_intact` |
| 6 | Org Owners/Admins retain implicit Agent Owner authority | ✅ | `authorization.py` `_scope` — Owner/Admin bypass not affected by this PR |
| 7 | `agent.access.manage` can read/set/change/remove general access | ✅ | `_require_manage_access` gates all endpoints; `test_member_without_access_manage_cannot_read_or_change_general_access` |
| 8 | Supports locked Viewer/Editor/Owner + same-org custom roles | ✅ | `test_owner_sets_changes_and_removes_general_access` (Viewer), `test_general_access_requires_role_granting_agent_read` (custom) |
| 9 | Only roles granting `agent.read` can be selected | ✅ | `_require_general_access_role` checks `AGENT_READ in permissions`; `test_general_access_requires_role_granting_agent_read` |
| 10 | Backend exposes current setting; supports set/change/return to Restricted | ✅ | `GET /share` + `PUT /share` (null = Restricted); `test_owner_sets_changes_and_removes_general_access` |
| 11 | List/count/detail/effective/subordinate resolve union | ✅ | `test_general_access_makes_agent_visible_with_role_permissions`, `test_general_access_scopes_subordinate_agent_resources` |
| 12 | Visibility before count/order/pagination; invisible→404, visible-missing-op→403 | ✅ | `test_unassigned_member_cannot_discover_general_access` (404), `test_general_access_makes_agent_visible_with_role_permissions` (403 for update) |
| 13 | DB + app validation prevent cross-org roles | ✅ | DB trigger + `_require_access_role` (org-scoped); `test_general_access_rejects_foreign_custom_role`, `test_general_access_rejects_custom_role_from_another_organization` |
| 14 | Referenced role cannot be deleted | ✅ | `ON DELETE RESTRICT`; `test_general_access_role_rejects_referenced_role_deletion` |
| 15 | Changes take effect on next request without reauthentication | ✅ | `test_direct_and_general_access_permissions_are_additive`, `test_share_snapshot_role_change_takes_effect_on_next_request` |
| 16 | Migration/auth/tenant-isolation/pagination/subordinate/membership-lifecycle/union tests | ✅ | 16 tests in `test_agent_general_access.py` + 5 in `test_rbac_schema.py` + updated `test_agent_rbac.py` |

---
*Reviewed by Hermes Agent*

api/domains/agents/repository.py
```diff
@@ -142,11 +153,116 @@ def find_agent_permissions(
...
+                return False
+
+            desired_role_ids = set(assignment_roles.values())
+            if general_access_role_id is not None:
+                desired_role_ids.add(general_access_role_id)
+            if desired_role_ids:
+                existing_role_ids = set(
+                    session.exec(
+                        select(AgentAccessRole.id)
+                        .where(col(AgentAccessRole.id).in_(desired_role_ids))
```
aai-labs-agent: 💡 **Defense-in-depth: role existence check is unscoped**

This query checks that role IDs exist globally without filtering by `organization_id`. The service layer has already validated org-scoping (`_require_access_role` scopes by org), and the DB trigger blocks cross-org writes, so this isn't exploitable. But adding an org filter would make the query's intent clearer and strengthen defense-in-depth:

```python
.where(
    col(AgentAccessRole.id).in_(desired_role_ids),
    or_(
        col(AgentAccessRole.is_system).is_(True),
        col(AgentAccessRole.organization_id) == organization_id,
    ),
)
```

*Surfaced during comparison with a colleague's review.*

api/domains/rbac/policy.py
```diff
@@ -15,6 +15,7 @@ class AuthorizationScope:
     organization_id: UUID
     membership_id: UUID | None = None
     permission: PermissionKey | None = None
+    include_general_access: bool = True
```
aai-labs-agent: 💡 **`include_general_access` defaults to `True` — consider `False` for fail-closed**

The default `True` is safe because the only code path that sets `membership_id` (the prerequisite for general access to take effect) is `AgentAuthorization._scope()`, which always explicitly sets `include_general_access`. But a default of `False` would be more conservative — if a future caller sets `membership_id` without setting this flag, they'd get direct-access-only rather than accidentally including general access.

*Surfaced during comparison with a colleague's review.*

api/domains/agents/access_service.py
```diff
@@ -42,8 +43,74 @@ def list_roles(self, context: CurrentUserContext) -> list[AgentAccessRoleRead]:
...
+        return AgentAccessSettingsRead(
+            general_access=AgentGeneralAccessRead(role=general_role_read),
+            assignments=assignments,
+        )
+
+    def _require_general_access_role(self, role_id: UUID | None, organization_id: UUID) -> AgentAccessRoleRead | None:
+        if role_id is None:
+            return None
+        role, role_read = self._require_access_role(role_id, organization_id)
+        permissions = self.rbac_repository.get_agent_access_role_permissions(role.id)
```
aai-labs-agent: 💡 **Redundant permission lookup**

`_require_access_role` (called at line 95) already fetches the role's permissions via `get_agent_access_role_permissions` and builds an `AgentAccessRoleRead` with them. But `_require_general_access_role` immediately re-fetches permissions at line 96 instead of using `role_read.permissions`:

```python
role, role_read = self._require_access_role(role_id, organization_id)
permissions = self.rbac_repository.get_agent_access_role_permissions(role.id)  # redundant
if PermissionKey.AGENT_READ not in permissions:
```

Could be:
```python
role, role_read = self._require_access_role(role_id, organization_id)
if PermissionKey.AGENT_READ not in role_read.permissions:
```

Minor inefficiency (extra DB round-trip), not a bug.

*Surfaced during comparison with a colleague's review.*

api/migrations/versions/dbec52536f66_merge_heads.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from typing import Sequence, Union
+
+# revision identifiers, used by Alembic.
+revision: str = "dbec52536f66"
+down_revision: Union[str, Sequence[str], None] = ("b4e7a21c9f35", "dce8d344498a")
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
```
dominykas-aai-labs, line 16: Rechaining after merging main would be cleaner maybe? But this works too I guess

PR #88, opened 2026-07-23
Af 172 per org model allowlist

miglmigl:
### Description
This PR implements the per-organization model allowlist feature (AF-172). Organizations can now restrict which AI models are available to their agents, giving administrators more control over model usage and access.

### What's Changed
* **Database Updates:** Added the `allowed_models` field to the `Organization` model and created an Alembic migration (`181dcfcc93ef`) for it.
* **Data Migration:** Backfilled existing organizations with `["*"]` so current users don't lose access to any models.
* **Service Logic:** Updated the `AgentService` and `OrganizationService` to actively filter available model catalogs against the organization's allowlist.
* **Tests:** Added comprehensive unit and integration tests (`test_agent_allowlist.py`, `test_organization_update_models.py`, `test_migration_backfill_allowed_models.py`) to verify the allowlist enforcement and migration logic.

### Notes
* **UI Bug:** Conversations are currently not displaying in the UI for agents (it shows "No conversations yet" even when messages should be present).


aai-labs-agent (changes requested):
## Code Review Summary

**Verdict: Changes Requested 🔴** (2 critical, 5 warnings, 5 suggestions)

**PR:** #88 — Af 172 per org model allowlist
**Author:** @miglmigl
**Files changed:** 24 (+1173 -72)

---

### 🔴 Critical

1. **Migration backfill semantic inversion** (`migration.py:196`) — The migration backfills `allowed_models` from the old `AGENT_MODEL_ALLOWLIST` config. But the semantics flipped: old empty allowlist = "allow everything"; new empty allowlist = "block everything." If the config is empty/unset in the deployment running this migration, every existing org is backfilled with `[]` — **locking out all models**. The PR description says backfill with `["*"]`, but the code doesn't do that. The migration test only covers a non-empty config value and misses this case.

2. **`ensure_default_organization` creates org with empty allowlist** (`organizations/service.py:126`) — `Organization(name="default", is_default=True)` doesn't set `allowed_models`, so it defaults to `[]` (blocks everything). New deployments that create a default org would have no models available. Should default to `["*"]` or the configured default model.

### ⚠️ Warnings

3. **`start_agent` doesn't enforce allowlist** (`agents/service.py:873`) — `_ensure_model_allowed` is only called on create/update, but the allowlist can change afterward. An agent whose model was valid at creation but later removed from the org's allowlist can still be started. The docstring's reasoning ("create/update are the only paths that can set an agent's model") doesn't account for the allowlist itself changing.

4. **`_validate_allowed_models` swallows all exceptions** (`organizations/service.py:61`) — `except Exception:` catches everything (network timeouts, parse errors, auth failures) and silently skips validation with no logging. A transient OpenRouter outage during an allowlist update would silently accept invalid patterns.

5. **`check_db.py` debug script committed** — Standalone script that creates a DB engine and prints org data. Appears to be a throwaway debugging artifact.

6. **UI `isDirty` false positive** (`allowed-models-section.tsx:83`) — The dirty check doesn't account for the auto-added default model, so the Save button is enabled on load when the default isn't in the DB allowlist.

7. **UI silently discards orphaned models** (`allowed-models-section.tsx:20`) — `getPrefixedModels` drops DB models not in the catalog. Users could unknowingly remove models from their allowlist on save.

### 💡 Suggestions

8. **SQL injection risk in migration** (`migration.py:198`) — f-string interpolation into raw SQL. Input is config-controlled, but a pattern with a single quote breaks the statement. Use parameterized execution.
9. **Migration includes unrelated noise** — Dozens of `alter_column(TEXT → AutoString)` calls unrelated to the feature. Trim to only `allowed_models` changes.
10. **Unrelated changes mixed in** — `cloud_id` fields on JiraContent/ConfluenceContent, `except` syntax fix in conversations/service.py, openclaw config pruning — should be separate PRs.
11. **`alert()` for error handling** (`use-organization-actions.ts:71`) — Blocks main thread; consider a toast notification.
12. **`_validate_allowed_models` O(n×m) loop** (`organizations/service.py:70`) — `catalog_ids_lower` recomputed per pattern; hoist above the loop.

### ✅ Looks Good
- Clean migration from comma-separated string to list-based allowlist with glob matching
- Good unit test coverage for `is_model_allowed` / `filter_models_by_allowlist`
- Proper authorization gating on the `?catalog=true` endpoint (`ORGANIZATION_UPDATE`)
- Prefix-stripping logic is well-tested (create + fetch round-trip)
- Migration test validates backfill behavior (just missing the empty-config edge case)
- `flag_modified` + single-session commit for JSONB mutation is the correct SQLAlchemy pattern

### Missing test coverage
- No test for `ensure_default_organization` with the new allowlist semantics
- No test for `_ensure_model_allowed` rejecting when org allowlist is `[]`
- No test for `start_agent` with a model not in the allowlist
- Migration test doesn't cover the empty-config case (the critical bug above)

---
*Reviewed by Hermes Agent*

aai-labs-agent:
## Re-Review: All Prior Findings Resolved ✅

**PR:** #88 — Af 172 per org model allowlist (updated)
**Verdict: Approved ✅** — all critical and warning-level issues from the first review have been addressed.

> This supersedes the earlier formal review. The inline comments from that review have been cleaned up since all findings were resolved.

---

### Previously Flagged → Now Fixed

| # | Finding | Status |
|---|---------|--------|
| 🔴 1 | Migration backfill defaults to `[]` when config empty (locks out all models) | ✅ Fixed — now defaults to `["*"]` with a clear comment explaining the semantic inversion |
| 🔴 2 | `ensure_default_organization` creates org with empty allowlist | ✅ Fixed — now sets `allowed_models=[default_model]` + unit test added |
| ⚠️ 3 | `start_agent` doesn't enforce allowlist | ✅ Fixed — re-checks allowlist at start time with a clear comment on why |
| ⚠️ 4 | `_validate_allowed_models` swallows all exceptions silently | ✅ Fixed — now catches `Exception as e` and logs via `logger.warning` |
| ⚠️ 5 | `check_db.py` debug script committed | ✅ Removed |
| ⚠️ 6 | UI `isDirty` false positive on load | ✅ Fixed — `getEffectiveModels` used for both initial state and dirty check |
| ⚠️ 7 | UI silently discards orphaned models on save | ✅ Fixed — orphaned models now preserved with amber "Not in catalog" badges and explicit remove control |
| 💡 8 | SQL injection risk in migration backfill | ✅ Fixed — now uses parameterized `sa.text()` with `bindparams()` |
| 💡 9 | `_validate_allowed_models` O(n×m) loop | ✅ Fixed — `catalog_ids_lower` hoisted above the loop |
| 💡 10 | `alert()` for error handling | ✅ Fixed — replaced with `toastError()` |
| 💡 11 | Migration includes unrelated `alter_column` noise | ⚠️ Noted — existing autogen artifacts, not blocking |
| 💡 12 | Unrelated changes mixed in (`cloud_id`, `except` fix, openclaw pruning) | ⚠️ Noted — would prefer separate PRs, but not blocking |

### New Test Coverage Added
- `test_ensure_default_organization_sets_default_model_allowlist` — verifies default org gets the configured model
- E2E test fixtures updated with `allowed_models: ["*"]` across all test data support files

### One Minor Nit (Non-blocking)
The `_ensure_model_allowed` docstring still says "create/update are the only paths that can set an agent's model; enforcing here is sufficient" — but `start_agent` now also enforces the check. The docstring could be updated to reflect the three enforcement points (create, update, start).

### ✅ Looks Good
- The orphaned-models UX is well thought out — amber badges with tooltip, explicit remove, and preservation on save
- `getEffectiveModels` / `getOrphanedModels` / `getPrefixedModels` decomposition is clean and prevents the dirty-check drift
- Parameterized SQL in the migration is the correct approach
- The `start_agent` allowlist re-check with a descriptive error message ("no longer in the organization's allowed model list") gives clear feedback
- Test coverage for the edge cases that were missing in the first revision

Ship it 🚀

---
*Re-reviewed by Hermes Agent*

aai-labs-agent:
## Code Review: Per-Org Model Allowlist (AF-172)

**Verdict: Changes Requested** — 2 warnings, 3 suggestions.

### Jira AC Cross-Check

| 1 | Admins have a UI to pick which models their org's agents can use | ✅ | `AllowedModelsSection` component with checkbox UI + search |
| 2 | The selected set is stored per-org | ✅ | `allowed_models` JSONB column on `Organization` + migration `181dcfcc93ef` |
| 3 | Model picker only shows org's allowed models | ✅ | `list_models(context, catalog=False)` filters via `filter_models_by_allowlist` using `organization.allowed_models` |
| 4 | Agent create/update rejects models outside the org's allowed set | ✅ | `_ensure_model_allowed(model, org_id)` fetches org and calls `is_model_allowed` |
| 5 | Existing `is_model_allowed` / `filter_models_by_allowlist` logic is reused | ✅ | Same functions, signature changed from `str` to `list[str]` — glob matching preserved |
| 6 | A new org starts with only the default model allowed | ✅ | `create_organization` defaults to `[config.agent_default_model]`; `ensure_default_organization` sets `[default_model]` |

### ⚠️ Warnings

1. **Server doesn't enforce that the default model must remain in `allowed_models`** (`organizations/service.py:59`) — The UI prevents removing the default model (checkbox `disabled={model.isDefault}`), but the API has no such guard. An admin could PATCH via API to remove the default. Agents using `config.agent_default_model` would pass create but fail at `start_agent` with a confusing error.

2. **`update_organization` with `allowed_models: null` writes NULL to a non-nullable column** (`organizations/service.py:208`) — When `allowed_models` is explicitly `None`, the code calls `flag_modified` and `setattr(organization, "allowed_models", None)`, writing NULL to a JSONB column with `server_default="[]"`.

### 💡 Suggestions

3. **Response key casing change is a potential breaking API change** (`agents/service.py:1446`) — `context_length` → `contextLength` and `is_default` → `isDefault`. The API client factory has `transformKeys: true`, so verify the transform is a no-op for already-camelCase keys.

4. **Stale docstring** (`agents/service.py:1421`) — Still says `is_default` but code now uses `isDefault`.

5. **Unrelated changes mixed in** (`agents/models.py:9`) — `cloud_id` fields, openclaw `tools.exec` pruning, and `except` syntax fix are unrelated to the allowlist feature.

### ✅ Looks Good

- **Start-time allowlist re-check** (`agents/service.py:875`) — correct defense-in-depth pattern; org's allowlist can change after agent creation.
- **Migration backfill handles empty-config semantic inversion** — `["*"]` backfill for empty old config. Both scenarios tested.
- **Orphaned model handling in the UI** — preserves models no longer in catalog instead of silently dropping.
- **Good test coverage** — unit, integration, and migration backfill tests.
- **Fail-closed semantics** — empty allowlist blocks everything.

- Migration auto-generated `alter_column` noise (~30 TEXT→AutoString) — standard Alembic behavior.
- `_validate_allowed_models` skips validation when OpenRouter unavailable — fail-open for validation only, not enforcement. Reasonable.
- `OrganizationRead.allowed_models: list[str] = []` — mutable default, safe in Pydantic v2.

---
*Reviewed by Hermes Agent*

api/domains/organizations/service.py
```diff
@@ -51,13 +56,48 @@ def get_organization(self, organization_id: UUID, context: CurrentUserContext) -
             )
         return organization
 
+    def _validate_allowed_models(self, allowed_models: list[str]) -> None:
```
aai-labs-agent: ⚠️ **Warning: Server doesn't enforce that the default model stays in `allowed_models`**

The UI prevents removing the default model (checkbox is `disabled={model.isDefault}`, `toggleModel` returns early for defaults), but the API has no equivalent guard here in `_validate_allowed_models`.

An admin could PATCH `{"allowed_models": ["some-other-model"]}` via API, removing the default model from the allowlist. Agents using `config.agent_default_model` (when no explicit model is set) would pass `create_agent` (model is None → skips `_ensure_model_allowed`) but then **fail at `start_agent`** with a confusing "no longer in the organization's allowed model list" error.

I'd add a server-side check: if the default model slug is not in the submitted list, either reject or auto-inject it — matching the UI's behavior.

api/domains/organizations/service.py
```diff
@@ -154,11 +196,27 @@ def update_organization(
...
 
-        organization_read = self.organization_repository.get_read(organization.id)
+        # Mutate and commit inside a single live session
+        # so SQLAlchemy properly tracks list mutations and flushes the UPDATE.
+        with Session(self.organization_repository.delegate.engine, expire_on_commit=False) as session:
+            session.add(organization)
+
+            from sqlalchemy.orm.attributes import flag_modified
+
+            if "allowed_models" in dump:
```
aai-labs-agent, line 206: ⚠️ **Warning: `allowed_models: null` writes NULL to a non-nullable column**

When `allowed_models` is explicitly set to `None` in the PATCH body, the code calls `flag_modified` and `setattr(organization, "allowed_models", None)` at line 215, writing NULL to the JSONB column.

The model declares `default_factory=list` and the migration sets `server_default="[]"` — the column should never be NULL. While `is_model_allowed` handles `None` as falsy (blocks everything), this is inconsistent with the schema's intent.

Consider converting `None` → `[]` here, or adding an explicit guard to skip the `setattr` when the value is `None`.

api/domains/agents/service.py
```diff
@@ -1400,40 +1417,53 @@ def _join_public_channels(self, bot_token: str, channel_ids: list[str]) -> None:
...
+            organization = self.organization_repository.get(org_id)
+            if not organization:
+                raise HTTPException(status_code=404, detail="Organization not found")
+            allowed = filter_models_by_allowlist(raw_catalog, organization.allowed_models)
         options = [
             {
                 "value": f"litellm/openrouter/{model['id']}",
                 "label": model["name"],
-                "context_length": model.get("context_length"),
+                "contextLength": model.get("context_length"),
```
aai-labs-agent, line 1604: 💡 **Suggestion: Response key casing change is a potential breaking API change**

The `list_models` endpoint changed response keys from `context_length` → `contextLength` and `is_default` → `isDefault`. The API client factory sets `transformKeys: true`, which likely converts snake_case → camelCase on the client side.

If the transform is a two-way conversion, sending camelCase from the backend while the client also transforms could cause issues. Any external API consumer expecting `context_length` would also break. Worth verifying the `transformKeys` behavior is a no-op for already-camelCase keys, or reverting to snake_case on the backend and letting the client handle the transform.

api/domains/agents/service.py
```diff
@@ -1400,40 +1417,53 @@ def _join_public_channels(self, bot_token: str, channel_ids: list[str]) -> None:
             except Exception as e:
                 logger.warning("Unexpected error joining channel %s: %s", channel_id, e)
 
-    def list_models(self, context: CurrentUserContext) -> list[dict]:
+    def list_models(self, context: CurrentUserContext, catalog: bool = False) -> list[dict]:
         """Returns the allowlisted OpenRouter models as picker options. The
```
aai-labs-agent, line 1579: 💡 **Suggestion: Stale docstring**

The docstring still says "flagged `is_default`" but the code now uses `isDefault` (line 1466). Minor, but could confuse someone reading the code.

api/domains/agents/models.py
```diff
@@ -118,12 +118,14 @@ class JiraContent(SecretContent):
     site_url: str
     email: str
     api_token: str
```
aai-labs-agent, line 124: 💡 **Suggestion: Unrelated changes mixed into the PR**

The `cloud_id` field on `JiraContent` and `ConfluenceContent` doesn't appear related to the per-org model allowlist feature. The openclaw `tools.exec` pruning and the `except (ValueError, TypeError)` syntax fix in `conversations/service.py` are also unrelated.

These are fine changes, but would be cleaner in separate commits or PRs for traceability — especially the `except` syntax fix, which is a bug fix that could get lost in a feature PR.

PR #91, opened 2026-07-27
AF-220 Event Delivery

kalibetre:
Introduce an internal domain event system built on a PostgreSQL transactional outbox and Dramatiq-based asynchronous delivery. Domain events are now staged within the same database transaction as the business write that produced them, persisted as immutable outbox messages, and fan-out into per-handler delivery records that progress through a durable lifecycle (pending → enqueued → processing → succeeded / dead-lettered). A reconciliation process and Kubernetes-deployed worker provide at-least-once delivery with bounded retries.

This establishes the foundation for internal event-driven projections (security audit, lifecycle notifications) without coupling producers to consumers or depending on a broker for durability.

## Transactional Outbox and Domain Event Contract

A new `events` domain introduces a typed, tenant-aware `DomainEventEnvelope` carrying actor identity, subject identity, correlation ID, and a bounded, secret-safe payload. A `DomainEventRegistry` validates every event against a registered definition: payload schema, sensitive-key scrubbing, organization-scoping checks, and a 16 KB size limit. Events are staged into the immutable `event_outbox_message` table via `OutboxMessageRepository.stage()`, which must be called inside the caller's open SQLAlchemy session so the outbox row commits atomically with the business change. A database trigger prevents updates to outbox rows, enforcing immutability.

For each staged event, the registry's declared handler names determine how many `event_delivery` rows are created — one per intended handler. Deliveries are the mutable, retryable unit of work; the outbox message itself never changes.

## Event Delivery Pipeline

Delivery processing follows a claim-based lifecycle:

- **Immediate dispatch**: `EventDeliveryDispatcher.enqueue_immediate()` publishes newly created delivery IDs to the Dramatiq Redis broker and marks them enqueued, called by services after transaction commit.
- **Worker processing**: `EventDeliveryProcessor` claims an enqueued delivery (or reclaims a stale processing one), validates the handler exists and supports the event, executes the handler, and records the outcome — succeeded, retryable failure (re-raises for Dramatiq retry), or terminal dead-letter.
- **Reconciliation**: `EventDeliveryReconciler` is a cronjob-runnable process that atomically claims stale pending, stale enqueued, and stale processing deliveries via `SELECT ... FOR UPDATE SKIP LOCKED`, then republishes them. Claiming and marking enqueued happen in the same transaction to eliminate the race between concurrent reconcilers.
- **Retry exhaustion**: A Dramatiq `on_retry_exhausted` callback dead-letters the delivery with `RETRY_EXHAUSTED`.

Structured logging covers handler duration, dead-letter events with reasons, and reconciliation summaries including pending count and oldest pending age.

## Domain Event Emission

Event emission is integrated into existing write paths across two domains:

- **Agents**: `agent.created` on creation, `agent.started` / `agent.stopped` on lifecycle transitions, and `agent.access.granted` / `agent.access.revoked` / `agent.general_access.changed` within the atomic access-settings snapshot replacement. All events share a correlation ID per access-settings operation.
- **Organization users**: `organization.role.changed` on member role change, staged atomically with the membership update under a row lock.

Each emission point resolves the actor identity from the current user context (membership preferred, falling back to user) and returns delivery IDs so the service layer can enqueue immediate dispatch after commit.

## Event Handlers

Two handlers are registered:

- **`SecurityAuditProjection`**: A placeholder sink for RBAC audit events (`organization.role.changed`, `agent.access.*`, `agent.general_access.changed`). It currently no-ops so those deliveries reach `SUCCEEDED` rather than permanently dead-lettering with `UNKNOWN_HANDLER`. The actual audit projection is deferred.
- **`AgentLifecycleEmailHandler`**: Sends lifecycle notification emails to the agent creator and owners on `agent.started` / `agent.stopped`. Per-recipient idempotency is enforced via `agent_lifecycle_email_receipt` keyed by `(delivery_id, recipient_email)`, so a retry after partial failure only re-notifies recipients that failed last time.

## Actor Identity Resolution

A shared `resolve_actor_identity()` helper maps the current user context to a typed `ActorIdentity` (membership or user), used consistently across the agents service, agent access service, and organization users service. This ensures all emitted events carry a durable, typed principal rather than a raw user ID.

## Technical Details

- **Dependencies**: Added `dramatiq[redis]` and `redis` to the API.
- **Configuration**: New `event_delivery_redis_url` setting, wired through Helm values and secrets.
- **Migrations**: Three new migrations — `event_outbox_message` (with immutability trigger), `event_delivery` (with status and dead-letter-reason enums and a check constraint linking dead-letter reason to status), and `agent_lifecycle_email_receipt`.
- **Dependency injection**: `AppModule` now provides `EventHandlerRegistry`, `EventDeliveryTransport`, `EventDeliveryReconciler`, and `EventDeliveryProcessor` as singletons.
- **Deployment**: Helm templates added for an event-delivery worker `Deployment` and a reconciliation `CronJob`.
- **Review fixes**: The final commit addresses atomic reconciliation claiming, event dispatch dedup, handler idempotency for lifecycle emails, `SecurityAuditProjection` registration, and non-atomic `agent.created` staging.

## Breaking Changes

- `AgentRepository.create_with_creator_access` now requires an `actor` keyword argument and returns `AgentLifecycleEventResult` (agent + delivery IDs) instead of `Agent`.
- `AgentRepository` now requires an `OutboxMessageRepository` dependency injected.
- `OrganizationUserRepository` now requires an `OutboxMessageRepository` dependency and exposes `change_role_with_event()` replacing direct role mutation in the service layer.
- `AgentRepository.replace_access_settings` accepts optional `actor` and `correlation_id` parameters and stages RBAC audit events within the same transaction.


api/domains/events/repository.py
```diff
@@ -0,0 +1,293 @@
...
+        eligible_conditions = [status_column == EventDeliveryStatus.ENQUEUED]
+        if processing_stale_before is not None:
+            claimed_at_column = cast(Any, EventDelivery.claimed_at)
+            eligible_conditions.append(
+                sa.and_(
+                    status_column == EventDeliveryStatus.PROCESSING,
+                    claimed_at_column.is_not(None),
+                    claimed_at_column <= processing_stale_before,
+                )
+            )
```
sami-bre, line 149: `claim_delivery` picks up a delivery only if it's ENQUEUED or it's (PROCESSING and stale). but the processor marks failed deliveries PROCESSING (in `mark_delivery_retryable_failure`). As a result, exponential backoff retry won't pick them up till they're stale (15 mins). I think the simple fix is to set failed deliveries status ENQUEUED?

sami-bre (changes requested):
All looks good. Just one thing ... 
Mentioned an issue that potentially turns dramatiq retries into fixed 15 min wait. Pls check it out.

PR #92, opened 2026-07-27
AF-166: allow deleting templates

sami-bre:
## Summary
- Add `DELETE /api/v1/templates/{slug}`, purging every version of a template lineage
- Block deletion while any live agent is pinned to the lineage (409); soft-deleted agents are detached instead of blocking, which required making `Agent.template_slug`/`template_version` nullable
- Pre-defined templates remain undeletable (403)
- Delete action lives in the template drawer footer as a red "Delete" button next to "Edit template", disabled with a tooltip when pre-defined or in use
- Fix `list_template_versions` to report `in_use` (previously only `list_templates`/`get_template` did)

## Test plan
- [ ] `make check-api` / `make check-ui` pass
- [ ] `make test-api` passes (950+ tests, including new delete/in_use coverage)
- [ ] Manually verified in a local compose stack: create → delete unused custom template, disabled state + tooltip for pre-defined and in-use templates, confirmation dialog, list refresh after delete


dominykas-aai-labs:
Suggestions (non-blocking)

1. **Missing RBAC test for delete** — There's test_member_cannot_create_template and test_member_cannot_update_template but no equivalent for delete. The guard exists in the service code and works identically, but the test pattern is incomplete.
2. **No test for GET /{slug} returning in_use: true** — List and versions endpoints are covered, but the single-template GET endpoint isn't tested for this flag.
3. **Unrelated .gitignore changes** mixed into the branch — harmless but could have been a separate commit.

sami-bre:
@dominykas-aai-labs
Missing tests added

PR #94, opened 2026-07-28
AF-236 and AF-238 Replace superuser/default-org model with platform admin and explicit organization scoping

kalibetre:
This PR restructures how platform-level authority and organization context work across Agent Farm. The "superuser who transcends org boundaries" model is replaced by a **Platform Administrator** role with authority that is deliberately separate from Organization Membership. The implicit "default organization" fallback is removed, and all organization-scoped API routes now carry the organization ID in their URL path. Predefined templates are refactored into true global platform resources with a dedicated `platform_template` table and real referential integrity for agent template pins.

---

### Platform Administrator replaces Superuser

The `is_superuser` flag on `User` is renamed to `is_platform_admin`, and every authorization path that previously granted superusers implicit org-scoped authority is tightened. Platform administrators no longer receive synthesized `OWNER` memberships when targeting an organization they don't belong to — they must have a **real persisted Membership** for any org-scoped operation. This closes a cross-org escalation vector where a superuser could act as owner of any organization without a membership row.

A new `PlatformAdminService` provides a dedicated seam for platform-level authority checks, and a `require_platform_admin` dependency guard replaces the old `check_superuser` parameter on `get_current_user`. The `CurrentUserContext.has_org_role` family no longer short-circuits on the platform flag; role resolution is always against the target organization's persisted membership map.

**Affected areas:** auth utils/models, RBAC policy, agent authorization, organization service, user service/repository, UI auth schemas and guard components.

### Explicit Organization Context via URL Scoping

Organization scoping moves from the `X-Organization-Id` request header to an explicit `/organizations/{organization_id}` path prefix on every org-scoped router (agents, slack, conversations, costs, skills, templates, tool calls). Platform-level routes (user management, organization creation/listing) move to a `/platform/...` prefix. The global `_default_org_id` singleton and `set_default_org_id` startup hook are removed entirely.

Personal endpoints (`/auth/me`, Slack config token, Google OAuth) now request the user context with `require_organization=False` so they work before an organization is selected. The UI `OrganizationProvider` and API instance are updated to pass the active organization ID via path rather than header.

**Affected areas:** all org-scoped routers, auth utils, auth routes, integration routes, UI shared API layer and organization provider.

### Removal of the Default Organization

The `is_default` column is dropped from `Organization` via migration. The startup lifespan no longer calls `ensure_default_organization()` or creates an `OWNER` membership linking the platform admin to the default org. Organization deletion no longer special-cases a default org. The `OrganizationRead` schema and UI organization schemas drop the `is_default` field. Environment variables are renamed from `SUPER_USER_CREDENTIALS`/`SUPER_USER_FULL_NAME` to `PLATFORM_ADMIN_CREDENTIALS` (full name defaults to "Super User").

### Two-Table Template Model with Real Referential Integrity

Predefined templates are split out of the org-scoped `agent_template` table into a dedicated global **`platform_template`** table (no `organization_id`), mirroring the built-in `aai_cli` skill model. This is a two-migration refactor: the first makes predefined templates global within `agent_template` (nullable `organization_id`), and the second moves them into `platform_template` and restores strict org-scoping on `agent_template`.

**Schema changes:**
- `platform_template` table with `UNIQUE(template_slug, version)` and a `platform_template_skill` join table.
- `agent_template` regains `NOT NULL` on `organization_id` and gains `forked_from_platform_template_id` (nullable FK → `platform_template.id`) to record fork lineage.
- The `Agent` model drops loose `(template_slug, template_version)` pin columns and replaces them with two mutually-exclusive FKs: `platform_template_id` and `agent_template_id`, enforced by a `CHECK` constraint (`ck_agent_exactly_one_template_pin`).

**Behavioural changes:**
- `TemplateRepository` resolves templates across both tables (org first, then platform); an org fork at a higher version shadows the platform v1.
- Editing a platform predefined template forks it into `agent_template` at `version = platform_v + 1` with `forked_from` set, instead of mutating the global row.
- The seeder writes to `platform_template` once at startup; per-org seeding on organization creation/signup is removed.
- `SkillRepository` deletion guards and stale-ref cleanup now check both join tables.
- `AgentService` uses a `_set_pin()` helper to set the correct FK; `AgentRead` derives `template_slug`/`template_version` from the pinned row at read time.

### Technical Details

- **Migrations:** `c6d7e8f9a0b1` drops `is_default`; `d5a6b7c8d9e0` makes predefined templates global within `agent_template`; `e6b7c8d9e0f1` splits them into `platform_template` and re-points agent pins via data migration. The `8b9c0d1e2` migration renames `is_superuser` to `is_platform_admin`.
- **`TemplateRead`** now has `organization_id: UUID | None` and `forked_from_platform_template_id: UUID | None`, reflecting that a template response may represent a global platform resource.
- **Bulk agent reads** are optimized: `get_pinned_template_info_for_agents` and `get_required_skill_ids_for_agents` batch-resolve across both pin kinds in a single pass, replacing per-agent slug/version lookups.
- **UI** adds a `/dashboard/platform` section (organizations, users) for platform admin view, separate from org-scoped dashboard pages. The org switcher and top nav are updated for the explicit-organization model.

### Breaking Changes

- **Environment variables:** `SUPER_USER_CREDENTIALS` → `PLATFORM_ADMIN_CREDENTIALS`; `SUPER_USER_FULL_NAME` removed (defaults to "Super User").
- **API routes:** All org-scoped endpoints now require `/organizations/{organization_id}` in the path instead of the `X-Organization-Id` header. User management and organization creation/listing moved to `/platform/users` and `/platform/organizations`.
- **Agent pin columns:** `agent.template_slug` and `agent.template_version` are dropped; agents now pin via `platform_template_id` or `agent_template_id` FK.
- **`UserRead.is_superuser`** renamed to `is_platform_admin`.
- **`OrganizationRead.is_default`** removed.
- **`TemplateRead.organization_id`** is now nullable; `forked_from_platform_template_id` added.


AnaniyaT (commented):
**`helmfile.yaml.gotmpl:137`**

Chart wants `platformAdminCredentials` now, the helmfile still sends `superUserCredentials`

**`compose.yml:32`**

Old name here too, so the API won't start: `platform_admin_credentials Field required`.

**`helm/agentfarm-api/Chart.yaml`**

Needs a bump for the versions, chart files changed and so did the API image.

.env.deploy.spec
```diff
@@ -59,7 +59,6 @@ SECRET_SIGNING_KEY=
 AGENT_TOKEN_ENCRYPTION_KEY=
 # Format: email:password
 SUPER_USER_CREDENTIALS=admin@example.com:
```
AnaniyaT: And here.

api/migrations/versions/e6b7c8d9e0f1_split_predefined_templates.py
```diff
@@ -0,0 +1,331 @@
...
+        sa.text(
+            """
+            UPDATE agent a
+            SET template_slug = pt.template_slug, template_version = pt.version
+            FROM platform_template pt
+            WHERE a.platform_template_id = pt.id
+            """
+        )
+    )
+    op.alter_column("agent", "template_slug", existing_type=sa.String(length=255), nullable=False)
```
AnaniyaT: These make the columns mandatory on rollback, but they were optional before this PR. Breaks on any install where a template lineage was deleted — the soft-deleted agents that used it have blank pins, so rollback dies: `NotNullViolation: column "template_slug" of relation "agent" contains null values`. Clean installs roll back fine, so it won't show up unless you test with that data.

api/domains/templates/service.py
```diff
@@ -199,72 +178,76 @@ def update_template(self, slug: str, data: TemplateUpdate, context: CurrentUserC
...
+        return self._to_read_with_skills(new_template)
 
     def delete_template(self, slug: str, context: CurrentUserContext) -> None:
         org_id = self._org_id(context)
         latest = self._get_latest_or_404(org_id, slug)
         self.permission_policy.require_organization(context, org_id, PermissionKey.TEMPLATE_MANAGE)
+        if not isinstance(latest, AgentTemplate):
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN,
+                detail="Cannot delete platform templates",
```
AnaniyaT, line 197: An edited copy keeps `PRE_DEFINED` and this blocks deleting those, so once an org edits a built-in template they're stuck with their copy and can't get back to the original — which keeps getting updated without them. Follow-up is fine, just flagging it.

api/domains/templates/repository.py
```diff
@@ -56,164 +107,302 @@ def find_versions(self, org_id: UUID, slug: str) -> list[AgentTemplate]:
...
+        return org_latest if org_latest.version >= platform_latest.version else platform_latest
+
+    def resolve_versions(self, org_id: UUID, slug: str) -> list[AgentTemplate | PlatformTemplate]:
+        org_versions = self.find_org_versions(org_id, slug)
+        platform_versions = self.find_platform_versions(slug)
+        combined: list[AgentTemplate | PlatformTemplate] = list(org_versions) + list(platform_versions)
+        combined.sort(key=lambda t: t.version, reverse=True)
+        return combined
+
+    def find_latest_templates(
```
AnaniyaT, line 301: Paging moved out of SQL into memory, this loads every template before slicing.

docs/architecture/api.md
```diff
@@ -24,9 +24,9 @@ Nearby domains are the implementation template. Costs and Ingest intentionally d
 
 ## Tenancy and authorization
 
-Authentication builds `CurrentUserContext`; organization-scoped services derive the active organization from it. `X-Organization-Id` selects the active organization, with the configured default organization as fallback. Normal users require membership. Tenant resolution synthesizes owner-level organization context for superusers, and authorization helpers explicitly preserve the superuser bypass.
+Authentication builds `CurrentUserContext`; organization-scoped services derive the active organization from it. Organization-scoped routes carry the organization ID in the URL, generally as `/api/v1/organizations/{organization_id}/...`. Routes without an `organization_id` path parameter have no active Organization. Normal users require membership. Tenant resolution synthesizes owner-level organization context for Platform Administrators only when they make an explicit org-scoped request, and platform-only routes use the platform-admin dependency without resolving an active Organization. Platform administration routes live under `/api/v1/platform/...`.
```
AnaniyaT: Still describes platform admins getting owner access to any org. That's what this PR removes.

api/domains/agents/service.py
```diff
@@ -622,21 +634,26 @@ def get_agent(self, agent_id: UUID, context: CurrentUserContext) -> AgentRead:
...
+                detail=f"Agent {agent_id} has no pinned template",
+            )
+        template = self.template_repository.resolve_template(org_id, pinned.template_slug, version)
         if not template:
             raise HTTPException(
                 status_code=status.HTTP_404_NOT_FOUND,
                 detail=f"Template version {version} not found for agent {agent_id}",
             )
-        return TemplateRead.model_validate(template)
+        return self.template_repository._to_read(template, self.template_repository.get_required_skills_for(template))
```
AnaniyaT: `_to_read` is private and in another domain.

.env.deploy.spec
kalibetre: Fixed — renamed `SUPER_USER_CREDENTIALS`/`superUserCredentials` to `PLATFORM_ADMIN_CREDENTIALS`/`platformAdminCredentials` in `.env.deploy.spec`, `compose.yml`, and `helmfile.yaml.gotmpl` to match the chart's `secret.yaml`. Also caught and fixed the same mismatch in `.github/workflows/deploy.yml`

api/migrations/versions/e6b7c8d9e0f1_split_predefined_templates.py
kalibetre: Fixed — removed the `NOT NULL` re-tightening on `agent.template_slug`/`template_version` in `downgrade()`. They're left nullable, matching the state before this migration (they were already loosened to nullable by e7a4b9c2d5f1 for exactly this soft-deleted/detached-pin case).

Verified by reproducing it directly: spun up a throwaway Postgres DB, migrated to head, inserted a soft-deleted agent with a detached (NULL) pin, then ran `alembic downgrade` — it failed with the exact `NotNullViolation` you described before the fix, and now completes cleanly with the row intact.

api/domains/templates/service.py
kalibetre: Agreed, and given the "follow-up is fine" I'm leaving this as-is for now rather than making a rushed call here. Fixing it isn't a small tweak — it also touches what happens to any live agents currently pinned to the org's forked version when that fork is deleted (the fallback would presumably resolve back to the platform original via `resolve_template`), and there's an existing test (`test_delete_predefined_template_returns_403`) asserting today's behavior that would need to change with it. We will do a proper pass at this with a follow-up ticket

api/domains/templates/repository.py
kalibetre: Fixed — `find_latest_templates` no longer loads full template rows (including the markdown body columns) to do the merge/sort/paginate step. The org and platform "latest per slug" queries now select only identity columns (id, slug, name, version, source), which is what the merge/shadow/sort logic actually needs; full rows (with the markdown bodies) are fetched afterward only for the ids on the resulting page.

docs/architecture/api.md
kalibetre: Fixed — reworded that paragraph. It now states that org-scoped access always requires a real persisted Membership and Platform Administrators get no synthesized/implicit organization role (confirmed against `auth/utils.py::get_authenticated_user`, which raises 403 if the caller has no Membership in the requested org, regardless of `is_platform_admin`).

api/domains/agents/service.py
kalibetre: Fixed — renamed `TemplateRepository._to_read` to a public `to_read` and updated both call sites (this one and `TemplateService._to_read_with_skills`), since it's a legitimate cross-domain read used by more than just the owning repository now.

AnaniyaT (approved):

PR #99, opened 2026-07-28
Af 165 make GitHub or bitbucket skill required on pr reviewer agent creation

sami-bre:
## AF-165: Require GitHub or Bitbucket for the PR Reviewer agent

- Adds a general "at least one of" (OR-group) requirement to the template required-skills model — a nullable `group_key` on `agent_template_skill` (`NULL` = existing AND-required behavior, shared key = pick-one-or-more group).
- PR Reviewer now seeds GitHub + Bitbucket under one group; agent creation is rejected (400) unless at least one is included.
- Backend: migration, group-aware template CRUD/seeding, agent create/update/list validation, and "required" stamping (a group member becomes required only once it's the sole assigned member).
- Hire dialog: renders each group as a multi-select "choose at least one" section, gates the Hire button on a pick, and shows credential forms only for the chosen member(s) — selecting both GitHub and Bitbucket is supported.
- Template editor: full authoring UI for groups — create a group from 2+ selected skills, add/remove members, dissolve a group back to standalone; view-mode preview also renders groups correctly.
- Tests: backend (templates + agents integration), UI e2e (hire-dialog + settings-templates) — all passing.


aai-labs-agent:
# AF-165: Require GitHub or Bitbucket for the PR Reviewer agent

**Reviewed by:** Code-Reviewer agent
**Jira:** AF-165 — Make GitHub or Bitbucket skill required on PR Reviewer Agent Creation
**Verdict: Comment** (1 suggestion, 1 documentation sync gap)

---

## Summary

This PR adds a general "at least one of" (OR-group) requirement model to template required skills via a nullable `group_key` column on `agent_template_skill`. The PR Reviewer predefined template now requires GitHub *or* Bitbucket instead of only GitHub. The implementation spans the full stack: migration, backend validation (create/update/list/seed), hire dialog, template editor, and comprehensive tests (backend integration + UI e2e).

The design is clean and well-reasoned. The `requirements.py` helper module is a good extraction — pure functions that are easy to test and reuse. The grandfathering logic for agents whose pinned template gains a group after creation is thoughtful and tested. The `_validate_no_skill_overlap` guard prevents the confusing state where a skill is both standalone-required and a group member. The partial-seed test (only one host skill exists) is a great edge case to cover.

## Acceptance Criteria Cross-Check

| 1 | PR Reviewer template must require either GitHub or Bitbucket | ✅ | `predefined/__init__.py`: `required_skill_names=(("GitHub", "Bitbucket"),)` — seeded as a group with key `github-or-bitbucket`. Test: `test_seed_predefined_templates_code_reviewer_requires_github_or_bitbucket` |
| 2 | Must not require only GitHub during agent creation | ✅ | `agents/service.py` `create_agent`: checks `split_requirements(required_map)`, rejects 400 if no group member is in `skill_ids`. Test: `test_create_agent_with_no_group_member_returns_400` |

## Findings

### 💡 Suggestions

- **`config-drawer.tsx` re-pin uses radio (single-choice) while hire dialog uses toggle (multi-choice)** — The hire dialog allows selecting *both* GitHub and Bitbucket (toggle/checkbox behavior in `toggleGroupMember`), and the backend supports it (`test_create_agent_with_both_group_members_marks_neither_required`). But the re-pin flow in `config-drawer.tsx` uses `<input type="radio">` which limits the user to exactly one member per group. This is a UX inconsistency between the two surfaces. If single-choice is intentional for re-pin (simpler), that's a reasonable design call — but worth confirming it's deliberate, since a user who had both GitHub and Bitbucket assigned and then re-pins to the same template would be forced to drop one.

### 📄 Documentation sync gap

- **`docs/features/templates-and-skills.md` not updated** — The repo's `AGENTS.md` review protocol (rule 3) requires: "A changed invariant, boundary, state model, or operational contract requires the authoritative document to change in the same diff." This PR introduces a new requirement model ("at least one of" groups) that changes the following documented invariants:
  - *"Template-required skills must be explicitly present on the agent"* (Skill invariants) — now it's "at least one of" for group members, not "all must be present."
  - *"Updating a custom template ... preserves required skills unless replacements are supplied"* (Template invariants) — now `required_skill_groups` is a new field that's also inherited/preserved.
  - *"Predefined template seeding may refresh ... required-skill associations in place"* — seeding now handles group-keyed entries and partial groups.
  - The Change-impact section should mention the new `group_key` column and the OR-group enforcement in agent create/update.

  `docs/features/templates-and-skills.md` should be updated in this diff to document the group model, the `group_key` column, and the updated invariants.

## Suppressed (checked, not flagged)

- **Migration has no backfill** — The migration adds `group_key` as nullable with no data backfill. This is correct: `NULL` means standalone AND-required, which matches all pre-existing rows. No backfill needed.
- **`save_template_skills` uses `session.add(row)` on existing rows** — In `repository.py`, when a row's `group_key` changes, the code calls `session.add(row)` on an already-persistent object. In SQLAlchemy this is fine (the session marks it dirty on flush), though the conventional pattern is to just mutate the attribute. Matches no specific repo convention I could find, and it works — not flagging.
- **`generateGroupKey` collision suffix could exceed 100-char backend limit** — The base key is sliced to 90 chars, then a numeric suffix is appended on collision (`${base}-${n}`). With enough collisions, `base(90) + "-999" = 94` chars, still under 100. Only flag if 10+ collisions on the same base are plausible, which is unlikely for a 2-3 member group. Not flagging.
- **`existing_skill_ids` fetched even for unrelated updates (grandfathered agents)** — In `update_agent`, the DB query for `existing_skill_ids` runs inside `if required_map:` but outside the `if data.removed_skill_ids:` and repin blocks. For a grandfathered agent making a rename-only update, this is one extra query. Minor and correct for the grandfathering logic — not flagging.
- **`splitRequiredSkills` folds 1-member groups into standalone in the UI** — A group with one member (e.g., partial seed with only GitHub) is displayed as a standalone required skill in the hire dialog, while the backend stores it with a `group_key`. Both paths require the skill, so behavior is consistent. Well-documented in the function comment — not flagging.

## ✅ Looks Good

- `requirements.py` is a clean, well-documented extraction — pure functions, no side effects, easy to test.
- Grandfathering logic in `update_agent` (only blocks group removal if the group *had* an assigned member before) is thoughtful and prevents breaking pre-existing agents when a template is reseeded.
- The `_validate_no_skill_overlap` guard in `models.py` prevents the confusing state of a skill being both standalone and in a group.
- Test coverage is thorough: create with no/one/both members, remove one/last member, swap in one call, repin with/without member, list marks sole member required, grandfathered agent — all covered.
- Seeding idempotency test (`test_seed_predefined_templates_is_idempotent_for_group_keys`) confirms re-seeds don't duplicate rows.
- Partial-seed test (only one host skill exists) covers the realistic deployment where Bitbucket isn't installed yet.
- UI e2e tests cover the hire-dialog group flow (choose, deselect, credential sync, request body) and the template-editor group authoring flow (create, add member, remove member, dissolve).

---
*Reviewed by Code-Reviewer agent*

ui/src/features/agents/components/config-drawer.tsx
```diff
@@ -675,6 +703,47 @@ export function ConfigDrawer({ agent, activeTab, onTabChange, onClose }: ConfigD
...
+                          (p) => !existingSecretProviders.has(p),
+                        );
+                        return (
+                          <label
+                            key={member.id}
+                            className="flex items-center gap-2 px-3.5 py-2.5 rounded-2xl text-[0.8125rem] cursor-pointer"
+                            style={{ border: "1px solid var(--line)", background: "var(--bg-soft)" }}
+                          >
+                            <input
+                              type="radio"
```
aai-labs-agent: 💡 **Suggestion** — Re-pin uses radio buttons (single-choice) here, while the hire dialog (`hire-dialog-steps.tsx`) uses toggle/checkbox behavior allowing *both* GitHub and Bitbucket to be selected. The backend supports both (`test_create_agent_with_both_group_members_marks_neither_required`).

If single-choice is intentional for re-pin (simpler), that's a reasonable design call — just confirming it's deliberate, since a user who had both hosts assigned and re-pins to the same template would be forced to drop one.

sami-bre: Confirmed not intentional — fixed in a40f4a3: the re-pin picker in `config-drawer.tsx` now uses checkboxes with the same multi-select semantics as the hire dialog, so re-pinning no longer forces dropping an already-assigned group member.

AnaniyaT (approved):

PR #100, opened 2026-07-29
Af 240 gracefully fail on predictable errors

dominykas-aai-labs:


api/domains/agents/scripts/hermes/healthz-server.py
```diff
@@ -135,4 +146,94 @@ def _send(self, code: int, body: dict) -> None:
...
+                        self.send_header(key, val)
+                self.end_headers()
+                while True:
+                    chunk = upstream.read(8192)
+                    if not chunk:
+                        break
+                    self.wfile.write(chunk)
+
+            conn.close()
+        except Exception:
```
aai-labs-agent, line 264: ⚠️ **Upstream connection leak + 502 attempted after headers sent**

`conn.close()` (line 223) sits inside the `try` block, after the streaming `while` loop. If `self.wfile.write(chunk)` raises (e.g. `BrokenPipeError` from client disconnect mid-stream), execution jumps here, skipping `conn.close()` — the upstream `HTTPConnection` leaks.

Additionally, the `self.send_response(502)` call below tries to write a new HTTP status line, but `end_headers()` was already called in the streaming branch (line 216). This writes a second response into an already-started one, corrupting the socket.

Suggested fix:
```python
_headers_sent = False
try:
    ...
    self.end_headers()
    _headers_sent = True
    while True:
        chunk = upstream.read(8192)
        if not chunk: break
        self.wfile.write(chunk)
finally:
    conn.close()

# in except:
if not _headers_sent:
    self.send_response(502)
    ...
```

aai-labs-agent (commented):
## Code Review — AF-240: Gracefully fail on predictable errors

**Author:** @dominykas-aai-labs | **Branch:** `AF-240-gracefully-fail-on-predictable-errors` → `main` | **Files:** 5 (+228 / −15)

**Verdict: Changes Requested** — 3 warnings, 3 suggestions. The proxy approach is sound and well-wired across all agent platforms, but the Python proxy has a connection-leak/error-handling issue in its streaming path, the Node.js proxy doesn't abort upstream on client disconnect, and the core feature (error interception) has no behavioral tests.

---

### What this PR does

Adds a sidecar LLM proxy (port 8090) inside both the Hermes (`healthz-server.py`) and OpenClaw (`healthz-server.js`) healthz servers. The proxy intercepts HTTP 401/402/403 responses from the LiteLLM upstream and replaces the raw error body with a concise, human-readable message. Non-terminal responses are forwarded unchanged. `service.py` routes all agent LLM traffic through `localhost:8090` and passes the real LiteLLM URL via `LITELLM_PROXY_TARGET`.

---

### ⚠️ Warnings

1. **Python proxy — upstream connection leaked + 502 attempted after headers sent** (`healthz-server.py:211–230`)
   `conn.close()` sits inside the `try` block after the streaming `while` loop. If `self.wfile.write(chunk)` raises (client disconnect during streaming), execution jumps to `except Exception`, skipping `conn.close()` — the upstream `HTTPConnection` leaks. The except block then calls `self.send_response(502)`, but `end_headers()` was already called in the streaming branch, so this writes a second HTTP status line into a response that's already in progress — corrupting the socket. Suggested fix: use `try/finally` for `conn.close()`, and guard the 502 path with a flag that tracks whether headers were already sent.

2. **Node.js proxy — client disconnect doesn't abort upstream request** (`healthz-server.js:193–236`)
   There's no `clientReq.on('close', …)` / `clientReq.on('aborted', …)` handler. If the client disconnects mid-LLM-call (agent restart, pod eviction, browser navigation for Teams), `upstreamReq` keeps running and the upstream connection leaks. Should call `upstreamReq.destroy()` on client close.

3. **No behavioral tests for the proxy's core feature** (`test_agents.py`)
   The two new tests verify that `LITELLM_PROXY_TARGET` and `LITELLM_BASE_URL` land in the right ConfigMap/Secret fields — good wiring coverage. But the actual proxy behavior (401→clean message, 402→clean message, non-terminal forwarding, 502 on upstream failure) is completely untested. A unit test that starts the proxy against a mock upstream returning 401/402/403 and asserts the clean body would close the gap.

### 💡 Suggestions

1. **402 message is OpenRouter-specific** (`healthz-server.py:30`, `healthz-server.js:23`)
   "OpenRouter credits exhausted. Add credits at https://openrouter.ai/credits." is correct for OpenRouter but misleading if LiteLLM routes to OpenAI/Anthropic directly. Consider making this provider-agnostic or deriving the message from the upstream error body.

2. **Hardcoded `http://localhost:8090` in service.py** (`service.py:916`)
   The proxy URL and `PROXY_PORT` are duplicated across `service.py` and both healthz-server scripts. If the port changes, three files need editing. Consider deriving from a config value or a shared constant.

3. **`assert` in production proxy code** (`healthz-server.py:176`)
   `assert _target_parsed is not None` — asserts are stripped under `python -O`. The proxy server only starts when `LITELLM_PROXY_TARGET` is set, so this is defensive code that's effectively dead. Replacing with an `if` guard + `send_error(500)` would be more robust.

---

### Acceptance Criteria Cross-Check

| 1 | Terminal errors (402, 401, invalid key, quota exceeded) detected, fail immediately, no retries | ⚠️ Partial | Proxy intercepts 401/402/403 and returns clean messages **with the same status code**. But no agent retry logic is modified in this PR — if the agent currently retries on 402 (per Jira root cause), the proxy alone doesn't change retry behavior. "Quota exceeded" may surface as 429, which the proxy forwards unchanged. |
| 2 | Concise notification: error type, one-line description, remediation link | ✅ Met | Clean messages include type, description, and link (e.g., `https://openrouter.ai/credits`). |
| 3 | Transient errors (429 w/ retry-after, 500/502/503) continue to retry | ✅ Met | Non-terminal status codes fall through to the `else` branch and are forwarded verbatim. |
| 4 | Applies to all agents and cron jobs | ⚠️ Partial | `service.py` covers all agent paths (Slack/Terminal/Teams × Hermes/OpenClaw). Cron job coverage unclear — cron jobs may use a separate code path that doesn't go through `start_agent`. |
| 5 | No nested `previous_errors` in user-facing notification | ⚠️ Not addressed | The proxy provides clean error messages, but the `previous_errors` concatenation is in the agent's error/notification handler, which is not modified in this PR. |

---

### ✅ Looks Good
- Proxy is feature-flagged via `LITELLM_PROXY_TARGET` — zero-impact when unset.
- Consistent error message format across both Python and JS implementations.
- `appVersion` bumped 0.32.1 → 0.32.2 (correct for code changes).
- Test coverage for the config/secret wiring is solid.
- SIGTERM cleanup added for the proxy server in the Node.js version.
- `ThreadingHTTPServer` used for the Python proxy (concurrent requests).

### Notes
- **Merge conflicts:** PR is in `dirty` mergeable state — needs rebase before merge.
- **PR description is empty** — consider adding a summary of the proxy approach and its limitations.

---
*Reviewed by Hermes Agent*

api/domains/agents/scripts/openclaw/healthz-server.js
```diff
@@ -165,6 +174,67 @@ const server = http.createServer((req, res) => {
...
+        error: { message: 'LLM proxy upstream unreachable', type: null, param: null, code: '502' }
+      });
+      clientRes.writeHead(502, {
+        'Content-Type': 'application/json',
+        'Content-Length': Buffer.byteLength(body),
+      });
+      clientRes.end(body);
+    });
+
+    clientReq.pipe(upstreamReq);
```
aai-labs-agent, line 253: ⚠️ **Client disconnect doesn't abort the upstream request**

There's no `clientReq.on('close', …)` / `clientReq.on('aborted', …)` handler. If the client disconnects mid-LLM-call (agent restart, pod eviction), `upstreamReq` keeps running and the upstream connection leaks.

Suggested addition:
```javascript
clientReq.on('close', () => upstreamReq.destroy());
```

Also consider destroying `clientReq` in the `upstreamReq.on('error')` handler to stop piping into a dead upstream.

api/domains/agents/scripts/hermes/healthz-server.py
```diff
@@ -20,6 +25,12 @@
 _TELEGRAM_API = "https://api.telegram.org"
 _SKIP_VALIDATION = os.environ.get("SKIP_SLACK_TOKEN_VALIDATION", "").lower() in ("1", "true", "yes")
 
+_TERMINAL_LLM_ERRORS: dict[int, str] = {
+    401: "LLM API key is invalid or expired. Check your API key configuration.",
+    402: "OpenRouter credits exhausted. Add credits at https://openrouter.ai/credits.",
```
aai-labs-agent, line 30: 💡 **402 message is OpenRouter-specific**

"OpenRouter credits exhausted. Add credits at https://openrouter.ai/credits." is correct when LiteLLM routes to OpenRouter, but misleading if the upstream is OpenAI, Anthropic, etc. directly. Consider making this provider-agnostic (e.g. "LLM billing credits exhausted. Check your provider billing page.") or deriving the message from the upstream error body.

api/domains/agents/scripts/hermes/healthz-server.py
```diff
@@ -135,4 +146,94 @@ def _send(self, code: int, body: dict) -> None:
...
+        self._forward()
+
+    def do_DELETE(self) -> None:
+        self._forward()
+
+    def do_PATCH(self) -> None:
+        self._forward()
+
+    def _forward(self) -> None:
+        assert _target_parsed is not None
```
aai-labs-agent: 💡 **`assert` in production code**

Asserts are stripped under `python -O`. The proxy server only starts when `LITELLM_PROXY_TARGET` is set, so this is effectively dead defensive code — but if someone runs the process with `-O` and the env var is somehow unset at request time, the handler will raise `AssertionError` with no clean error response. An `if not _target_parsed: self.send_error(500); return` guard would be more robust.

api/domains/agents/service.py
```diff
@@ -913,6 +913,7 @@ def start_agent(self, agent_id: UUID, context: CurrentUserContext) -> AgentRead:
             else ""
         )
         effective_model = agent.model or self.config.agent_default_model
+        llm_proxy_url = "http://localhost:8090"
```
aai-labs-agent, line 1034: 💡 **Hardcoded proxy URL**

`http://localhost:8090` is duplicated here and as `PROXY_PORT = 8090` in both healthz-server scripts. If the port needs to change, three files must be edited in sync. Consider deriving this from a config value or a shared constant.

AnaniyaT (approved):

PR #104, opened 2026-08-02
AF-237 Introduce self-service organizations and platform oversight

kalibetre:
This PR shifts organization provisioning to authenticated users while narrowing platform administration to essential onboarding and privilege management. It adds dedicated, allowlisted Platform Oversight views for users, organizations, memberships, and searchable member lists, with matching dashboard detail pages and clearer navigation.

### Organization and User Administration

- Allow users, including Platform Administrators, to create organizations for themselves.
- Persist immutable Organization Creator provenance and enforce a configurable per-creator creation limit.
- Simplify organization creation and platform user onboarding around invitations.
- Replace platform password-reset and account-deletion actions with Platform Privilege grant/revoke and resend-invite operations.
- Prevent self-revocation and removal of the final Platform Administrator.
- Standardize dashboard page width and improve grid card actions and navigation.

### Cross-Cutting Event and Audit Architecture

- Add explicit Platform and Organization event scopes with validation ensuring organization references match the event boundary.
- Introduce durable, idempotent Security Audit Records containing actor and subject display snapshots.
- Preserve outbox, delivery, and audit history when organizations or related entities are deleted.
- Restrict platform authority to user-session credentials, preventing API keys, service credentials, and runtime credentials from accessing platform administration.

### Technical Details

Platform Oversight uses dedicated read models rather than organization-scoped DTOs or membership permissions, ensuring platform administrators can inspect allowlisted metadata without gaining organization access. Database migrations backfill Organization Creator values, introduce scoped event constraints, and create durable audit storage. API, integration, unit, and end-to-end coverage has been updated for the new lifecycle and authorization behavior.

### Breaking Changes

- Organization creation moves from the platform organization endpoint to authenticated self-service organization creation.
- Organization creation no longer accepts owner or model-provisioning fields.
- Platform user creation no longer accepts passwords, organization IDs, or roles.
- Platform reset-password and delete-user endpoints are removed and replaced with privilege and invitation endpoints.
- Platform read responses use dedicated oversight models and include creator and membership details.


AnaniyaT:
`TokenData.credential_class` defaults to `USER_SESSION` (`auth/models.py:37`) and none of the four call sites overrides it, so the gate at `auth/utils.py:153` can never fire. Decode has the same default (`auth/utils.py:62`), meaning a token with no claim is trusted as a user session — so the first API-key issuer that forgets to stamp it silently gets platform authority. Making the field required and denying on an absent claim would make this match the "default-deny" wording in the AC better.

PR #107, opened 2026-08-04
AF-172 follow-up: fix stale Hire Agent model list after allowlist changes

miglmigl:
Follow-up to AF-172. Hire Agent's model picker only reflected allowlist changes after a full page reload.

Root cause: useModels() caches for an hour, and nothing invalidated it when allowed_models changed. Fix: invalidate the models query whenever an org update includes allowedModels.

### DIFF
--- a/ui/src/features/organizations/hooks/use-organization-actions.ts
+++ b/ui/src/features/organizations/hooks/use-organization-actions.ts
@@ -2,6 +2,7 @@
 
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 
+import { agentsKey } from "@/features/agents/utils";
 import { api } from "@/shared/api";
 import { toastError } from "@/shared/toast";
 
@@ -66,6 +67,11 @@ export function useUpdateOrganization() {
       void queryClient.invalidateQueries({
         queryKey: organizationsKey.detail(variables.organizationId),
       });
+      // The Hire Agent model picker (useModels) caches its result for an hour;
+      // an allowlist change must be reflected immediately, not after a reload.
+      if (variables.data.allowedModels !== undefined) {
+        void queryClient.invalidateQueries({ queryKey: agentsKey.models() });
+      }
     },
     onError: (error: Error) => {
       toastError(error, "Failed to save changes. Please try again.");


aai-labs-agent:
## Code Review — PR #107

**Verdict: Approve with one suggestion** — the fix is correct and well-targeted.

### Summary

Follow-up to AF-172. `useModels()` caches for an hour (`staleTime: 60 * 60_000`), and nothing invalidated that cache when an org's `allowedModels` changed. This PR adds a targeted `invalidateQueries` call in `useUpdateOrganization`'s `onSuccess` so the model picker reflects allowlist changes immediately without a page reload.

### Verification

- **Query key prefix match**: `agentsKey.models()` returns `["agents", "models"]`. `useModels` uses `[...agentsKey.models(), { catalog }]` as its query key. `invalidateQueries({ queryKey: agentsKey.models() })` uses TanStack's prefix matching — both `catalog: true` and `catalog: false` variants are invalidated. ✅
- **Guard condition**: `variables.data.allowedModels !== undefined` correctly scopes the invalidation to updates that actually include the allowlist (name/description-only edits skip it). ✅
- **staleTime bypass**: `invalidateQueries` marks matching queries as stale and refetches active ones regardless of the 1-hour `staleTime`. ✅
- **Org-switch isolation**: `"agents"` is already in `ORG_SCOPED_QUERY_KEYS` in `OrganizationProvider`, so the models cache is dropped on org switch. This PR only needs to handle same-org allowlist changes. ✅
- **Cross-feature import**: Importing `agentsKey` from `@/features/agents/utils` into an org hook is a necessary dependency for cross-feature cache invalidation — the webapp guideline says "Mutations MUST invalidate affected list and detail families." ✅
- **Import ordering**: External → `@/*` → relative, consistent with `docs/guidelines/webapp.md`. ✅

### Suggestion

- **Regression test**: `docs/guidelines/testing.md` says "Regression fixes SHOULD include a test that fails for the original defect." A Playwright test that updates the org's allowed models and then opens the Hire Agent dialog to verify the picker reflects the new list (without a page reload) would prevent this from regressing. See inline comment.

### Notes

- **Merge conflicts**: The PR is currently in a `dirty` mergeable state — rebase against `main` before merging.
- No documentation update needed — this fixes a cache invalidation bug, not a documented invariant or boundary.

---
*Reviewed by Hermes Agent — AGENTS.md review protocol followed (mapped through `docs/INDEX.md` → `docs/guidelines/webapp.md` → `docs/architecture/ui.md`)*

ui/src/features/organizations/hooks/use-organization-actions.ts
```diff
@@ -66,6 +67,11 @@ export function useUpdateOrganization() {
       void queryClient.invalidateQueries({
         queryKey: organizationsKey.detail(variables.organizationId),
       });
+      // The Hire Agent model picker (useModels) caches its result for an hour;
+      // an allowlist change must be reflected immediately, not after a reload.
+      if (variables.data.allowedModels !== undefined) {
```
aai-labs-agent, line 72: **Suggestion — regression test**

`docs/guidelines/testing.md` says: "Regression fixes SHOULD include a test that fails for the original defect."

A Playwright test in `ui/tests/e2e/hire-dialog.spec.ts` (or `organizations.spec.ts`) that updates the org's `allowedModels` via the settings UI, then opens the Hire Agent dialog and asserts the model picker reflects the new list — without a page reload — would lock in this fix. The existing `hire-dialog.spec.ts` already mocks `GET /agents/models`; the regression test would intercept a second response with a different allowlist after the mutation succeeds.

PR #108, opened 2026-08-04
AF-234-telemetry-push-fix

dominykas-aai-labs:


hermes-base/smoke-test.sh
```diff
@@ -32,6 +32,25 @@ with sync_playwright() as p:
...
+from hermes_cli.plugins import VALID_HOOKS
+
+hooks = {'pre_gateway_dispatch', 'post_llm_call', 'pre_tool_call', 'post_tool_call', 'on_session_end'}
+assert hooks <= VALID_HOOKS, hooks - VALID_HOOKS
+assert 'tool_call_id' in _DEFAULT_PAYLOADS['pre_tool_call']
+assert 'tool_call_id' in _DEFAULT_PAYLOADS['post_tool_call']
+assert hasattr(SessionStore, 'list_sessions')
+fields = set(SessionEntry.__dataclass_fields__)
+assert {'session_id', 'session_key', 'origin', 'chat_type'} <= fields, fields
+"
```
kalibetre, line 82: **[Standards / Smell]** This check covers hook names, `tool_call_id` presence, and top-level `SessionEntry` fields (including `session_key`, which `_resolve_chat` never reads) — but never asserts `session_store` is in `_DEFAULT_PAYLOADS['pre_gateway_dispatch']` or `session_id` is in `_DEFAULT_PAYLOADS['post_llm_call']`, the two kwargs the whole chat-resolution path depends on. It also doesn't check that `origin` exposes `chat_id`/`thread_id` (read via `getattr` in `__init__.py:84,88`).

Per `docs/guidelines/testing.md`'s Runtime plugin tests section: "Anything that depends on runtime behavior MUST also be checked inside the pinned image."

**Fix**: assert `session_store`/`session_id` presence in `_DEFAULT_PAYLOADS`, and check `origin`'s sub-fields instead of the unused `session_key`.

openclaw-base/smoke-test.sh
```diff
@@ -32,6 +32,18 @@ with sync_playwright() as p:
...
+# runId, and pairs tool results through toolCallId. OpenClaw exposes hook names
+# only as types, so this asserts against the shipped declarations: enough to
+# catch a release that renames a hook or drops a correlation field.
+check telemetry-contract sh -c '
+    types="$(npm root -g)/openclaw/dist/plugin-sdk/hook-types-"*.d.ts
+    for token in message_received agent_end before_tool_call after_tool_call \
+                 session_end sessionKey runId toolCallId; do
+        grep -q "$token" $types || exit 1
+    done
+'
```
kalibetre, line 45: **[Standards / Smell]** This new `telemetry-contract` check has no matching trigger-path update. `.github/workflows/hermes-base.yml` added `api/domains/agents/scripts/hermes/plugins/**` to its trigger paths so its equivalent check reruns on plugin edits, not just version bumps — but `.github/workflows/openclaw-base.yml`'s `on.push.paths` wasn't given the same treatment for `api/domains/agents/scripts/openclaw/plugins/**`. This PR's own OpenClaw plugin edit won't trigger this check pre-merge.

**Fix**: add `api/domains/agents/scripts/openclaw/plugins/**` (and the openclaw driver fixture path) to `openclaw-base.yml`'s trigger paths.

api/domains/agents/scripts/openclaw/plugins/telemetry-push/index.js
```diff
@@ -96,8 +97,26 @@ function extractText(msg) {
...
+function rememberChat(key, record) {
+  if (chatContexts.size >= MAX_CHAT_CONTEXTS) {
+    chatContexts.delete(chatContexts.keys().next().value);
+  }
+  chatContexts.set(key, record);
+}
+
+function toolKey(event, ctx) {
+  return `${event.runId || (ctx && ctx.runId) || ""}:${event.toolName || ""}`;
+}
```
kalibetre, line 120: **[Standards / Smell]** `toolKey(event, ctx)` keys only on `${runId}:${toolName}` with no per-invocation discriminator. When `event.toolCallId` is absent, two concurrent calls to the same tool within one run still collide last-writer-wins in `pendingToolCalls` (used at line 204 and lines 221-223) — reproducing, in this fallback path, the exact race class this diff fixes everywhere else.

**Fix**: include a per-invocation discriminator in the fallback key (e.g. a queue/counter per `runId:toolName`) so overlapping same-tool calls don't overwrite each other's pending id.

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -58,6 +58,39 @@ def _now_iso():
...
+    for entry in entries:
+        if getattr(entry, "session_id", None) != session_id:
+            continue
+        origin = getattr(entry, "origin", None)
+        chat_id = str(getattr(origin, "chat_id", "") or "")
+        if not chat_id:
+            return None
+        chat_type = str(getattr(entry, "chat_type", "") or "dm").lower()
+        thread_id = str(getattr(origin, "thread_id", "") or "") or None
+        return chat_id, chat_type, thread_id
```
kalibetre, line 89: **[Standards / Smell]** Minor: `_resolve_chat` returns a bare `(chat_id, chat_type, thread_id)` tuple, unpacked positionally by its one caller — a Data Clump.

**Fix**: low priority for a single-caller private helper; a small `NamedTuple` would make the unpacking self-documenting if more callers appear.

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -202,13 +243,15 @@ def _on_post_llm_call(session_id=None, user_message=None, assistant_response=Non
...
     external_id = f"hermes:{task_id}:{tool_name}:{int(time.time() * 1000)}:{_next_counter()}"
 
-    with _tool_call_ids_lock:
-        key = f"{task_id}:{tool_name}"
-        _tool_call_ids[key] = external_id
+    # Keyed on the per-invocation id: concurrent calls to one tool within a task
+    # share task_id and tool_name, so those would collide.
+    if tool_call_id:
+        with _tool_call_ids_lock:
+            _tool_call_ids[tool_call_id] = external_id
```
kalibetre, line 254: **[Spec / Scope Creep]** The tool_call_id rekeying here (and in `_on_post_tool_call` at lines 270-277, plus the OpenClaw `pendingToolCalls` map in `index.js:117-119,198-235`) changes hook signatures and drops the old `task_id:tool_name` fallback keying, but isn't requested by any AF-234 acceptance criterion.

AF-234's title and all three ACs say "message"/"chat" only; tool-call/result pairing is a task-scoped race, not a chat-session one. Arguably licensed by the broadly-worded Root Cause note ("a race condition in the telemetry push path"), but not by the literal ACs.

Suggest calling this out explicitly in the PR description as a related-but-separate fix, so it's evaluated on its own correctness bar rather than assumed covered by AF-234.

api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py
```diff
@@ -121,6 +154,14 @@ def _flush_loop():
 
 
 def _on_pre_gateway_dispatch(event, **kwargs):
+    global _session_store
+
+    # Held for post_llm_call, which receives only a session_id.
+    session_store = kwargs.get("session_store")
+    if session_store is not None:
+        with _session_store_lock:
+            _session_store = session_store
```
kalibetre, line 163: **[Spec / Suspicious]** Nothing in this diff verifies the real Hermes gateway actually calls `pre_gateway_dispatch` with `session_store=` at runtime. Unit tests inject a fake store directly, `hermes_session_store_driver.py` supplies `session_store=store` itself rather than exercising the real gateway's dispatch code, and `hermes-base/smoke-test.sh`'s contract check never asserts `session_store` is part of `_DEFAULT_PAYLOADS['pre_gateway_dispatch']`.

Confirmed empirically: swapping in the pre-fix plugin code makes the concurrency tests fail as expected, but nothing exercises this specific runtime-contract assumption either before or after the fix.

Relevant AC: "Existing single-chat telemetry behaviour is unchanged." If this assumption is false, `_resolve_chat` always returns `None` and every outbound Hermes reply is silently dropped — breaking that AC outright with nothing shipped to catch it.

**Fix**: add a static assertion (see the `hermes-base/smoke-test.sh` comment above) or extend the pinned-image contract step to confirm the real gateway supplies `session_store=`/`session_id` to these hooks.

kalibetre (commented):
## Code Review: `staging` → `HEAD` (AF-234)

Two-axis review (Standards + Spec) against `docs/guidelines/{code,testing,operations}.md` and Jira [AF-234](https://aai-labs.atlassian.net/browse/AF-234).

| Axis | Hard | Judgement | Total |
|------|------|-----------|-------|
| Standards | 0 | 4 | 4 |
| Spec | 0 | 2 | 2 |

**Worst Standards issue**: `index.js`'s tool-call fallback keying (`toolKey`) reintroduces the exact same-tool concurrency race this diff fixes everywhere else.
**Worst Spec issue**: nothing verifies the real Hermes gateway supplies `session_store=` at runtime — if false, AC3 ("existing single-chat telemetry behaviour is unchanged") breaks silently with no test catching it.

Findings are left as inline comments below.

Separately confirmed empirically: swapping in the pre-fix plugin code makes 10 of the new/changed tests fail (including both concurrency-attribution tests for Hermes and OpenClaw), and all tests pass against the fixed code, including the two `test_ingest.py` integration tests against a real Postgres. The core fix is verified — the items below are about test/CI coverage completeness and scope, not the core fix's correctness.

PR #109, opened 2026-08-05
AF-209-add-slack-channel-data-access-to-aai-cli-for-agent-barn-agents

sami-bre:
## Summary
- Adds `bookmarks:read` to the gateway's Slack app manifest scopes so a single Slack app/bot token covers both the OpenClaw/Hermes gateway and aai-cli channel-data access.
- API now auto-mirrors the gateway bot token into the aai-cli `SecretProvider.SLACK` Agent Secret whenever the Slack skill is attached to a `platform=slack` agent — no manual token entry.
- Mirrored secret is removed when the Slack skill is removed, and re-synced automatically if the gateway bot token is rotated.
- Slack skill is now platform-gated to `platform=slack` agents (400 with a clear message otherwise); direct client writes/removals of the `slack` Agent Secret are rejected (400) since it's now server-managed.
- UI: removed Slack as a manually addable/removable integration; hire dialog, agent Skills tab, and config-drawer's template re-pin panel show an informational note instead of a token field, and disable the Slack skill tile with "Requires Slack platform" for non-Slack agents.


api/migrations/versions/d731eac8a160_add_slack_to_secret_provider.py
```diff
@@ -0,0 +1,31 @@
+"""Add slack to agent_secret provider check constraint.
+
+Revision ID: d731eac8a160
+Revises: 181dcfcc93ef
+"""
+
+from alembic import op
+
+revision: str = "d731eac8a160"
+down_revision: str | None = "181dcfcc93ef"
```
aai-labs-agent, line 10: 🔴 **Deploy blocker:** `down_revision = "181dcfcc93ef"` is **not the current head** on the `staging` branch. Three other migrations already branch from `181dcfcc93ef` (`2a4f6c8e1b30`, `b3f7c1d92a04`, `b7f3d8e1c4a9`), and staging currently has 7 heads. Adding this migration creates an 8th branch from the same revision.

The PR also shows `mergeable_state: "dirty"` (merge conflicts), which suggests the branch was cut from an older staging where `181dcfcc93ef` was the head. After rebasing onto current staging, update `down_revision` to point to the current head (or add a merge migration). Otherwise Alembic will fail at deploy with "Multiple heads are present; please specify which head to begin from."

api/domains/agents/service.py
```diff
@@ -877,6 +936,48 @@ def update_agent(self, agent_id: UUID, data: AgentUpdate, context: CurrentUserCo
         if data.skill_ids or data.removed_secret_providers:
             self._validate_skill_update(agent, data, org_id)
 
+        # Keep the aai-cli Slack Agent Secret in sync with the gateway bot token:
+        # write/refresh it whenever a remaining skill requires it (covers skill add
+        # and bot token rotation), delete it when no remaining skill needs it
+        # anymore (covers skill removal). Independent of the validation gate above
+        # since skill removal alone must still trigger this.
+        if agent.platform == AgentPlatform.SLACK and (
```
aai-labs-agent: 💡 **Transaction ordering:** This Slack secret sync commits (via `PostgresRepositoryDelegate` session-per-operation) *before* the integration-secret validation block at line 985. If a co-submitted integration secret fails validation (422), the mirrored Slack secret is already committed while the skill assignment (line 1015) hasn't been applied — leaving an orphaned Slack secret.

This self-heals on the next skill/bot-token update (the sync would see no remaining skill requires Slack and delete it), so it's low severity. But consider moving this block after the integration-secret validation (line 985+) or after skill changes (line 1015+) to avoid the intermediate inconsistent state.

aai-labs-agent (changes requested):
## Code Review Summary

**Verdict: Changes Requested** — 1 deploy blocker, 1 suggestion

### 🔴 Critical (Deploy Blocker)

- **`api/migrations/versions/d731eac8a160_add_slack_to_secret_provider.py:10`** — The migration's `down_revision` (`181dcfcc93ef`) is not the current head of the `staging` branch. Three other migrations (`2a4f6c8e1b30`, `b3f7c1d92a04`, `b7f3d8e1c4a9`) already chain from `181dcfcc93ef`, and the staging branch currently has 7 heads. Adding this migration creates an 8th branch from the same revision. The PR also shows `mergeable_state: "dirty"` (merge conflicts). At deploy time, Alembic will fail with "Multiple heads are present; please specify which head to begin from." **Rebase onto current staging and update `down_revision` to the current head, or add a merge migration.**

### 💡 Suggestion

- **`api/domains/agents/service.py:944`** — The Slack secret sync commits (via `PostgresRepositoryDelegate` session-per-operation) before the integration-secret validation block at line 985. If a co-submitted integration secret fails validation (422), the mirrored Slack secret is already committed while the skill assignment (line 1015) hasn't been applied yet — leaving an orphaned Slack secret. This self-heals on the next skill/bot-token update (the sync logic would see no remaining skill requires Slack and delete it), but it's a consistency gap. Consider moving the Slack secret sync after the integration-secret validation, or after skill changes are applied.

### ✅ Looks Good

- The auto-mirroring design is solid: `create_agent` mirrors the gateway bot token into the `SecretProvider.SLACK` Agent Secret when a resolved skill requires it; `update_agent` syncs on skill add, skill remove (deletes), and bot-token rotation (re-encrypts). All three lifecycle paths are covered.
- Platform gating is correctly enforced in both entry points: `_resolve_skills` (create) and `_validate_skill_update` (update) both reject the Slack skill on non-Slack agents with a clear 400 message.
- Direct client writes/removals of the Slack Agent Secret are correctly rejected (400) in both create and update paths — the server-managed invariant is enforced.
- The `_remaining_skills` helper extraction is clean — preserves the original behavior while making the logic reusable for the Slack sync.
- Test coverage is comprehensive: create mirrors, PATCH add mirrors, rotate resyncs, remove deletes, non-Slack platform rejected (create + PATCH), explicit Slack secret rejected (create + PATCH), plus unit tests for `SlackContent` validation, encrypt/decrypt round-trip, the `validate_slack` integration validator (valid/invalid/scope/network error), config TOML, setup.sh, env vars, and tool-context MD.
- The `SlackContent` model, migration (check constraint), config TOML profile block, setup.sh secret, env-var mapping, and integrations policy MD all follow the established patterns for other providers.
- UI changes correctly hide Slack from manual configuration across all three surfaces (hire dialog SkillsStep, agent SkillsTab, config-drawer repin panel) and show an informational note instead of a token field.
- `bookmarks:read` scope addition to `BOT_SCOPES` is minimal and correct.
- The skill doc (`aai_cli_skills/slack.py`) is thorough — covers all command groups, response shapes, error taxonomy, and the canvas_id extraction caveat.

- **`assert` for `slack_bot_token`/`slack_app_token` in `create_agent`** — pre-existing, not introduced by this PR. Under `python -O` the asserts are stripped, but downstream `validate_content` catches a `None` token via Pydantic. Matches the established repo pattern for all platform configs.
- **Hardcoded `"slack"` string comparisons in UI files** — matches the established pattern for other providers in the same files (e.g., `"github"`, `"jira"`).
- **`validate_slack` registered in `_VALIDATORS` despite Slack secrets being server-managed** — the `validate_integration` endpoint (service.py:1772) can still be called for `SecretProvider.SLACK` to validate the mirrored secret on demand. Not dead code.
- **Real workspace/channel IDs in the skill doc examples** — reference documentation for agents; IDs are not sensitive and make the docs concrete.

---
*Reviewed by Hermes Agent*

sami-bre:
Merged latest `staging` and addressed both review points:

**🔴 Deploy blocker (migration head fork)** — confirmed: `staging` had advanced to 7 heads merging into `ee8815d37c1c` since this branch was cut from `181dcfcc93ef`. Added merge migration `71a0d9a1a91d` reconciling `d731eac8a160` (this PR's migration) with `ee8815d37c1c`, following the repo's established merge-migration convention. `alembic heads` now returns a single head, and the full chain was verified to replay cleanly on a disposable scratch Postgres.

**💡 Transaction-ordering suggestion** — moved the Slack Agent Secret sync block in `update_agent` to run last, after skill changes are applied and after all other secret/shared-credential validation and persistence. A later validation failure (e.g. a co-submitted invalid integration secret) can no longer leave the mirrored Slack secret committed while the rest of the update aborts. The `create_agent` path already had equivalent ordering after the earlier fixes in this branch, and now the Slack-derive block there also runs strictly after the shared-credentials attach loop that `staging` introduced.

The merge itself also had to reconcile staging's new shared-credentials and OR-group required-skills features with this PR's platform-gating/auto-sync logic (`_resolve_skills`/`_validate_skill_update` now take both the Slack platform check and staging's `extra_providers`/shared-credential coverage; the Slack informational-block special-casing now sits alongside staging's shared `IntegrationFields`/`SharedManualToggle` components).

All checks green after the merge: `check-api`, `check-ui` clean; `test-api` 1200 passed; `test-ui` 199 passed.

api/migrations/versions/d731eac8a160_add_slack_to_secret_provider.py
aai-labs-agent, line 10: 🔴 **Critical: `down_revision` points to a non-head revision.**

`181dcfcc93ef` is NOT a current head on the staging branch. It's already chained from by three other migrations (`2a4f6c8e1b30`, `b3f7c1d92a04`, `b7f3d8e1c4a9`). This creates an additional branch in the Alembic graph off an interior node, not a head.

The `down_revision` should point to one of the 7 current heads on staging. Alembic will report multiple heads at deploy time with the current chain.

api/migrations/versions/71a0d9a1a91d_merge_heads.py
```diff
@@ -0,0 +1,23 @@
...
+Revises: d731eac8a160, ee8815d37c1c
+Create Date: 2026-08-05 07:48:09.783509
+
+"""
+
+from collections.abc import Sequence
+
+# revision identifiers, used by Alembic.
+revision: str = "71a0d9a1a91d"
+down_revision: str | Sequence[str] | None = ("d731eac8a160", "ee8815d37c1c")
```
aai-labs-agent, line 13: 🔴 **Critical: merge with a non-head revision.**

`ee8815d37c1c` is NOT a current head on staging — it's already chained from by `1aa6ed23c288` (which is itself a head). The staging branch already has 7 heads; this merge resolves none of them and adds an 8th.

The merge should resolve `d731eac8a160` with the actual head it branched from, and ideally also merge the remaining heads to produce a single head.

api/migrations/versions/d731eac8a160_add_slack_to_secret_provider.py
```diff
@@ -0,0 +1,31 @@
...
+branch_labels: str | None = None
+depends_on: str | None = None
+
+_OLD = (
```
aai-labs-agent, line 14: ⚠️ **`_OLD` is missing `'pipedrive'`.**

The migration `6c23680e1076_add_pipedrive_to_secret_provider.py` (present on staging, in the active chain via `96333a5d58e4` → `6c23680e1076` → `1aa6ed23c288`) added `'pipedrive'` to the `ck_agent_secret_provider` constraint. But this `_OLD` (used for downgrade) and `_NEW` (used for upgrade) both omit it.

The `upgrade()` silently drops `'pipedrive'` from the allowed providers. If Pipedrive is intentionally being removed (it's not in the `SecretProvider` enum), that's fine — but `_OLD` should still include `'pipedrive'` so the `downgrade()` recreates the constraint as it actually was before this migration ran.

api/domains/agents/service.py
```diff
@@ -1034,6 +1096,49 @@ def update_agent(self, agent_id: UUID, data: AgentUpdate, context: CurrentUserCo
         for skill_id in dict.fromkeys(data.skill_ids):
             self.repository.add_skill(agent.id, skill_id)
 
+        # Keep the aai-cli Slack Agent Secret in sync with the gateway bot token:
```
aai-labs-agent, line 1099: ⚠️ **Comment says "runs last" but `self.repository.save(agent)` is at line 1142, after this block.**

If the Slack sync fails (e.g. `delete_secret` raises), skill changes (committed at lines 1094–1097) are orphaned and the agent record is not updated. This matches the existing pattern for other secret/skill changes in the method, but the comment is misleading — the sync runs second-to-last, not last.

Consider moving `self.repository.save(agent)` before the Slack sync block, or updating the comment to say "runs after every other secret/skill change" rather than "runs last."

aai-labs-agent (changes requested):
## Code Review Summary

**Verdict: Changes Requested** (2 critical, 2 warnings, 1 suggestion)

### 🔴 Critical

- **Migration `d731eac8a160` `down_revision` points to a non-head revision** — `down_revision = "181dcfcc93ef"`, but `181dcfcc93ef` is NOT a current head on the staging branch. It's already chained from by three other migrations (`2a4f6c8e1b30`, `b3f7c1d92a04`, `b7f3d8e1c4a9`). This creates an additional branch in the Alembic graph off an interior node, not a head.

- **Merge migration `71a0d9a1a91d` merges with a non-head revision** — `down_revision = ("d731eac8a160", "ee8815d37c1c")`, but `ee8815d37c1c` is also NOT a head on staging — it's already chained from by `1aa6ed23c288`. The staging branch already has 7 heads; this PR adds an 8th without resolving the existing ones. Alembic will report multiple heads at deploy.

### ⚠️ Warnings

- **`_OLD` check constraint in `d731eac8a160` is missing `'pipedrive'`** — The migration `6c23680e1076_add_pipedrive_to_secret_provider.py` (present on staging, in the active chain) added `'pipedrive'` to the `ck_agent_secret_provider` constraint. But this PR's `_OLD` (used for downgrade) and `_NEW` (used for upgrade) both omit it. The `upgrade()` silently drops `'pipedrive'` from the allowed providers. If Pipedrive is intentionally being removed (it's not in the `SecretProvider` enum), that's fine — but the `_OLD` value should still match the actual database constraint for the downgrade path to be correct.

- **`update_agent` Slack sync runs before `self.repository.save(agent)`** — The comment at line 1099 says "runs last, after every other secret/skill validation and persistence step above" but `self.repository.save(agent)` is at line 1142, after the sync. If the sync fails (e.g. `delete_secret` raises), skill changes (committed at lines 1094–1097) are orphaned and the agent record is not updated. This matches the existing pattern for other secret/skill changes in the method, but the comment is misleading.

### 💡 Suggestions

- **`_remaining_skills(agent.id, [], [])` at line 1108** — Passing empty add/remove lists reads the current post-update state from the DB (since skill changes were already applied at 1094–1097). A one-line comment noting this intentional read of post-update state would help future readers understand why the lists are empty.

### ✅ Looks Good

- Clean refactor of `_remaining_skills` into a reusable helper — logic is identical, and the early-return on empty result is actually safer than the old inline check
- Comprehensive test coverage: create, update, remove, rotate, platform-gating, and explicit-secret-rejection paths all tested
- `_TOOL_CONTEXT_PROVIDERS` correctly excludes Slack (no per-secret metadata to surface) — matches the pattern of Gmail/Zoho Mail
- `_INTEGRATION_LABELS` and `PROFILE_SLUGS` correctly include Slack
- Integration validator correctly handles Slack's HTTP 200 + `{"ok": false}` error pattern with clear error-code mapping
- `bookmarks:read` scope addition to the manifest is minimal and correct
- UI changes consistently filter out Slack as a manually configurable integration across all three surfaces (hire dialog, skills tab, config drawer) with clear "uses existing bot token" messaging
- `_resolve_skills` signature change correctly adds `platform` parameter — only one caller (in `create_agent`), and it's updated
- Slack skill content file (`slack.py`) is thorough and well-documented

### Suppressed

- **`PROVIDER_VALIDATORS` missing `FIRECRAWL`/`ZOHO`/`GOOGLE_CALENDAR`**: matches the existing pattern — not all providers have validators. Slack correctly gets one.
- **`hire-dialog-steps.tsx` `platform` type is `"slack" | "telegram"`**: consistent with the existing hire dialog which only supports these two platforms (Teams has a separate flow).

---
*Reviewed by Hermes Agent*

PR #111, opened 2026-08-05
AF-167: extend Domain Event coverage for the Security Audit Record projection

sami-bre:
## Summary

AF-167 ("Add Audit Logging") was blocked while AF-219/AF-220 (Domain Event/Outbox foundation) and AF-237 (Platform admin views) shipped independently. AF-237 already built the durable `security_audit_record` table and a real `SecurityAuditProjection` handler, covering RBAC/platform-privilege events. A separate ticket, AF-249, now owns the read-side "unified audit explorer" UI/API.

This PR fills in AF-167's remaining scope: broadening which mutations produce a Domain Event that feeds the existing `SecurityAuditProjection` — no new table, no new UI, no new handler.

- 12 new events, all routed to the existing `security_audit.projection` handler:
  - `agent.updated`, `agent.deleted`
  - `agent.secret.added` / `.updated` / `.removed`
  - `template.created` / `.updated` / `.deleted` (template create/update now atomically stage their event alongside the required-skill sync, in one transaction)
  - `organization.model_allowlist.changed`
  - `organization.member.added` / `.removed`, `organization.ownership_transferred`
- Secret event payloads are built allowlist-style from safe fields only (`provider`, `label`, `shared_reference_id`) and never include the encrypted `content`; field names deliberately avoid `secret`/`credential` substrings since the registry's sensitive-key filter rejects those.
- `agent.updated`/`template.updated` carry a generic `field_changes` diff, scoped to the fields actually safe/useful to audit (agent: name/model/approval_mode/template pin; template: name/description only, excluding markdown bodies) — no event is staged when nothing tracked actually changed.
- Removed now-dead non-event repository methods along the way (`save_secret`, `delete_secret`, `save_org_template_skills`, plain `transfer_ownership`).


AF-167_plan.md
```diff
@@ -0,0 +1,305 @@
+# AF-167 — Extend Domain Event coverage for the Security Audit Record projection
```
aai-labs-agent: 💡 This 305-line internal implementation plan (Jira links, design decisions, development notes) is an artifact of the development process, not part of the codebase. I'd remove it from this commit or add `*_plan.md` to `.gitignore` to prevent future plan files from slipping in.

ui/src/features/event-deliveries/utils.ts
```diff
@@ -2,7 +2,7 @@ import { createQueryKeyStructure } from "@/shared/query-keys";
 
 import type { EventDeliverySortDirection, EventDeliveryStatus } from "./schemas";
 
-export const EVENT_DELIVERIES_PAGE_SIZE = 50;
+export const EVENT_DELIVERIES_PAGE_SIZE = 20;
```
aai-labs-agent, line 5: 💡 This page-size change (50→20) appears unrelated to AF-167's audit-event-coverage scope. The repo's AGENTS.md asks for no unrelated churn in a diff. If this is a deliberate tweak, it'd be cleaner in its own PR so the change history stays traceable.

aai-labs-agent (commented):
## Code Review — AF-167: Extend Domain Event coverage for the Security Audit Record projection

**Verdict: Comment** — no blockers. Two cleanup suggestions and one documentation gap.

---

### 💡 Suggestions

1. **`AF-167_plan.md` — internal plan document committed to repo root.** This 305-line implementation plan (Jira links, design decisions, development notes) is an internal artifact that shouldn't be in the repo. Remove it from this commit or add `*_plan.md` to `.gitignore`. (Inline comment on the file.)

2. **`ui/src/features/event-deliveries/utils.ts` — unrelated page-size change.** `EVENT_DELIVERIES_PAGE_SIZE` was changed from 50 to 20, but this PR's scope is audit event coverage, not the event deliveries monitor UI. The repo's AGENTS.md says "The diff contains no unrelated churn." If this is an intentional separate tweak, it belongs in its own PR. (Inline comment on the file.)

### 📝 Documentation Gap

3. **Epic CHANGELOG not updated.** Per the repo's AGENTS.md review protocol: "For a PR belonging to an active multi-PR epic, reviewers MUST also verify that `docs/features/<epic-slug>/CHANGELOG.md` records the delivered slice and resulting current state." This PR delivers AF-167 under the AF-218 epic, but `docs/features/domain-events/CHANGELOG.md` was not updated — it currently records AF-247 as the latest entry. A new section recording AF-167's 12 new events, their scope, and the resulting current state should be added in this diff.

---

### ✅ Looks Good

- **Secret payload safety** — payloads are built allowlist-style from safe fields only (`provider`, `label`, `shared_reference_id`), never include the encrypted `content`, field names deliberately avoid `secret`/`credential` substrings to pass the registry's sensitive-key filter, and the Pydantic model uses `extra="forbid"` so even a buggy caller can't smuggle `content` in. Unit tests explicitly verify both properties.
- **Atomic transactions** — template create/update now collapse `save_template` + `save_org_template_skills` into one session with the event, eliminating the two-commit window where skills could persist without the template or vice versa.
- **No-op suppression** — `agent.updated` and `template.updated` correctly stage no event when no tracked field actually changed, avoiding confusing empty-content audit rows. Tests cover the no-op case for both.
- **`replace_access_settings` return type fix** — changing `bool` → `list[UUID] | None` fixes a latent bug: the old `not result` check would raise a 404 when access settings succeeded but staged zero events (empty list is falsy). The new `is None` check correctly distinguishes "agent not found" from "success with no changes."
- **Comprehensive test coverage** — 29 new tests across unit (payload validation, secret rejection, handler wiring, access-service enqueue) and integration (event emission per event, no-op suppression, full outbox→handler→`security_audit_record` projection for representative events). The integration tests drive real HTTP + Postgres.
- **Documentation** — `docs/features/domain-events.md` extended with all 12 new events and their payload semantics.
- **Dead code cleanup** — `save_secret`, `delete_secret`, `save_org_template_skills`, and plain `transfer_ownership` removed, replaced by their event-emitting counterparts.

### Notes on suppressed candidates

- **`soft_delete_with_event` doesn't scope its SELECT by `organization_id`** (unlike `update_scalar_fields_with_event` which does). This matches the established `_record_agent_event` pattern used by `create_with_creator_access`/`start_agent`/`stop_agent`, which also select by `agent.id` only. Suppressed — consistent with repo convention, and the service layer has already authorized the agent's org membership.
- **`test_outbox_messages.py` adds one-transaction persistence tests for only 2 of 12 events** (member.removed, ownership.transferred). The other 10 events are covered by full HTTP integration tests in their domain-specific test files, which assert the outbox message is persisted — the same property, tested differently. Not a gap.

---
*Reviewed by Hermes Agent*

api/domains/agents/repository.py
```diff
@@ -513,6 +517,212 @@ def _record_agent_event(
...
+                    type=SubjectIdentityType.AGENT,
+                    id=persisted.id,
+                    organization_id=persisted.organization_id,
+                ),
+                correlation_id=correlation_id or uuid4(),
+                payload={
+                    "organization_id": persisted.organization_id,
+                    "agent_id": persisted.id,
+                    "field_changes": field_changes,
+                    "actor_display": actor.type.value,
```
AnaniyaT: **Blocking: `actor_display` records the actor _type_, not the acting user.**

`actor.type.value` here is the literal string `"MEMBERSHIP"`. `SecurityAuditProjection.handle` copies `payload["actor_display"]` straight into `SecurityAuditRecord.actor_display` — the human-readable "who did this" column — so every audit record these events produce will read **"MEMBERSHIP"** rather than a name. AF-167's AC is "Every entry records the acting user".

Every pre-existing event passes the real display (`context.user.full_name or context.user.email`) — see `access_service.py:92` and the four call sites in `organization_users/service.py`, which this PR gets right. The other eight don't:

- `api/domains/agents/repository.py` — 580, 621, 667, 717
- `api/domains/templates/repository.py` — 211, 265, 585
- `api/domains/organizations/service.py` — 248

The fix looks mechanical: thread an `actor_display: str | None = None` parameter down from the service the way `replace_access_settings` already does, and keep `actor.type.value` only as the fallback for the system actor.

api/domains/organizations/service.py
```diff
@@ -207,13 +219,47 @@ def update_organization(
...
+                    subject=SubjectIdentity(
+                        type=SubjectIdentityType.ORGANIZATION,
+                        id=organization_id,
+                        organization_id=organization_id,
+                    ),
+                    correlation_id=uuid4(),
+                    payload={
+                        "organization_id": organization_id,
+                        "previous_models": previous_models,
+                        "new_models": organization.allowed_models,
```
AnaniyaT: **Blocking: this can make an org's model allowlist permanently un-editable.**

The payload carries the full before _and_ after lists. `DomainEventRegistry._validate_payload` caps payloads at 16 KB (`MAX_PAYLOAD_BYTES`) and raises `DomainEventValidationError` — which is a `ValueError`, not an `HTTPException`, and nothing on this path catches it. So once an org's allowlist is big enough, `update_organization` 500s and the whole transaction rolls back: the admin can no longer change the allowlist at all, including shrinking it back to something that would fit.

I measured the cliff on this branch: it breaks at roughly **230 entries per list** with 30-char model ids. `allowed_models` has no length bound — it's validated against the OpenRouter catalog, which is well over 400 models. `test_organization_model_allowlist_changed_payload_handles_realistic_large_list` only proves 150+150 fits, so it passes right up to the edge without pinning it down.

Suggest staging the diff (`added` / `removed`) instead of both full lists — bounded by the size of the actual change rather than the size of the allowlist, and more useful to read back in an audit log than two 200-item arrays.

AnaniyaT (commented):
Two blocking findings from a pass over this branch at `0b7d74c`. Both are in the event payloads rather than the plumbing — the one-transaction staging pattern itself looks right, and reusing AF-237's projection instead of rebuilding one was the correct call.

I have a handful of non-blocking notes too (layering on the org-service staging, the repeated `select(EventDelivery.id)` re-query, the bundled Event Delivery Monitor pagination rewrite) — happy to add those separately if useful.

Verification on this head: `ruff check` + `ruff format --check` + `ty check` clean, 902 API tests pass (all of `tests/unit` plus the six touched integration modules), UI `tsc --noEmit` and `eslint` clean.

sami-bre:
Review comments addressed:

- **Blocking: missing `actor_display`** — fixed across all 8 event-emitting call sites (`update_scalar_fields_with_event`, `soft_delete_with_event`, `save_secret_with_event`, `delete_secret_with_event`, agent access grant/revoke, org allowlist update). Added regression assertions in `test_agents.py`, `test_agent_general_access.py`, and `test_organization_update_models.py`.
- **Blocking: unbounded allowlist payload** — `OrganizationModelAllowlistChangedPayload` now carries the `added`/`removed` diff instead of full before/after lists, avoiding the `MAX_PAYLOAD_BYTES` cliff once the org's allowlist grows large. Covered by `test_organization_model_allowlist_changed_payload_carries_a_diff_not_full_lists` (300+300 entries).
- **`AF-167_plan.md` in repo root** — removed, and `*_plan.md` added to `.gitignore` to prevent recurrence.
- **Missing actor/subject identities on `agent.access.*`** — added a write-time `member_display` snapshot (name/email of the target member) to `AgentAccessGrantedPayload`/`AgentAccessRevokedPayload`, resolved via a batched query in `AgentRepository.replace_access_settings` rather than joining at read time, consistent with the retention ADR (records must stay readable after the membership/user is deleted).

Also merged latest `staging` into the branch (template-key rename, Slack-secret-sync path, Alembic head split) and re-ran `make check-api`, `check-ui`, `check-migrations`, `test-api`, `test-ui` clean post-merge. CI is green.

PR #117, opened 2026-08-11
AF-94: add a local k8s (k3d) dev and CI environment

AnaniyaT:
Adds a local k3d cluster for developing the agent path, and fixes the wiring that stopped a locally-run API from actually managing agents in it.

## What's included

- **`docker/k3d/`** — `k3d-up.sh` creates the cluster from a helper container (no host `k3d`/`helm` install), writes host and in-container kubeconfigs to `.k3d/`, seeds the namespace and the `litellm` secret, and adds a CoreDNS alias for `host.docker.internal` on native Linux. `k3d-load-images.sh` builds and imports the agent base images.
- **`Makefile`** — `cluster-up` / `cluster-down` / `cluster-delete` / `cluster-reset` and `k3d-load-images` (plus per-runtime variants). `cluster-down` **stops** the cluster, matching `db-down`/`redis-down`; `cluster-delete` is the destructive one. `dev-api` now starts the ingest app alongside the main app, mirroring the container entrypoint, and takes an overridable `API_DEV_PORT` so a second worktree can run its own stack.
- **`compose.yml`** — publishes the ingest port and overrides the three values that must differ inside the container (`K8S_KUBECONFIG_PATH`, `INGEST_BASE_URL`, `LITELLM_BASE_URL`). Everything else arrives via the existing `env_file`.
- **Telemetry reachability** — agent pods were given the in-cluster ingest Service address, which only resolves when the API itself runs in k8s. Locally the API runs outside the cluster, so pods now come back through the host. Without this, agents ran and chatted normally while their activity silently never persisted.
- **`openclaw-base/` + `hermes-base/` Dockerfiles** — an `APT_MIRROR` build arg, defaulting to `deb.debian.org`, so **CI and existing builds are unchanged**. Lets a developer route around a degraded Debian CDN edge, which stalled base-image builds for over an hour and made apt start skipping packages.
- **`.github/workflows/api.yml`** — the k8s integration job runs on k3d (`AbsaOSS/k3d-action`, `k3d-version` pinned because the action's default now 404s on install).
- **`README.md` + `.env.spec`** — the full local setup: one table of the values the k3d flow needs, the org-creation step a fresh database requires before agents can be created, the telemetry hop with a troubleshooting probe, and the `APT_MIRROR` escape hatch. `.env.spec` previously seeded none of the k3d values, so `make setup` produced a `.env` that couldn't run `cluster-up`.
- **`api/tests/integration/test_agents.py`** — covers the ingest wiring written into the agent secret, which had no test.

## Required repo config

None. No new GitHub Variables or Secrets; `APT_MIRROR` is an optional build arg with a default that preserves current behaviour.

## Notes

- No chart version bump: no `helm/` files changed.
- Verified live end to end on a from-scratch cluster for **both runtimes** — agent created with real Slack tokens, deployed, and telemetry persisted — with the API run in Docker and again on the host.
- Windows is documented as WSL2 or Git Bash; neither was verified on a Windows machine. Note the Makefile already required a Unix shell before this change (`setup`, `merge-heads`, `makemigrations`, `check-migrations`, `check-monitoring`).
- `tool_call` telemetry is unexercised: the live agents were only exchanged greetings, which need no tools.

## Testing

`make check-api`, `make test-api` (1261 passed), `make check-migrations`, `make lint-ui`, `make check-ui`, `make check-monitoring`, `docker compose config`. Live: `cluster-up`, `cluster-down` (stop/start cycle preserving imported images and running agents), `k3d-load-images`, `k3d-load-hermes`, `migrate`, `up`, `dev-api`, `dev-ui`; agent create/start/stop/delete for OpenClaw and Hermes; pod→ingest→Postgres telemetry confirmed for both, plus plugin-emitted Slack conversation rows. Both base images pass the new smoke tests. CI green on `b18a6a7b`, including `ci-api / Test k8s integration`.


aai-labs-agent:
## Code Review — AF-94: local k8s (k3d) dev and CI environment

**Verdict: Comment** — 2 suggestions, nothing blocking. Solid PR.

### Summary

This is a well-structured infrastructure PR. The k3d dev environment, CI workflow migration from bastion SSH to k3d-action, APT_MIRROR escape hatch, and telemetry wiring fix are all clean. The README documentation is thorough and the new integration test for the ingest secret wiring fills a real gap. I verified the repo conventions and checked the existing `docs/features/` docs — no doc sync required for this change (the local-dev INGEST_BASE_URL override is a dev concern, not a production invariant change).

### 💡 Suggestions

- **`compose.yml:170`** — litellm-db volume mount path doesn't match the repo convention. The main `db` service mounts at `/var/lib/postgresql/data`; litellm-db mounts at `/var/lib/postgresql` (missing `/data`). Both work, but matching the existing pattern would be cleaner.

- **`docker/k3d/k3d-up.sh:188`** — the summary message says `make dev-api → also run 'make dev-ingest' in a second shell`, but `dev-api` now starts ingest automatically alongside the main app (the trap-based co-start in the Makefile). This message is stale and contradicts both the Makefile and the README's "starts ingest alongside the main app" description.

### ✅ Looks Good

- **CI migration** — replacing the bastion SSH tunnel with `AbsaOSS/k3d-action` makes CI self-contained and removes the external k8s cluster dependency. The k3d version is pinned with a "keep in sync" comment.
- **`dev-api` trap pattern** — `trap 'kill 0' EXIT INT TERM` correctly cleans up the background ingest process. Good call starting ingest automatically to prevent the silent-no-persistence failure mode.
- **Integration test** — `test_start_agent_wires_telemetry_push_into_the_secret` verifies the `INGEST_URL` and `INGEST_API_KEY` are wired into the agent secret. Fills a real gap — this path had no test.
- **APT_MIRROR build arg** — defaults to `deb.debian.org` (CI unchanged), handles both classic and deb822 source layouts, and the rationale is well-documented.
- **kubeconfig dual-variant approach** — writing host (127.0.0.1) and internal (host.docker.internal) kubeconfigs from a single fetch is clean.
- **CoreDNS host alias** — the native Linux `host.docker.internal` resolution via `coredns-custom` is the right approach.
- **LiteLLM config** — wildcard passthrough with `store_model_in_db: false` is appropriate for local dev.

### Notes

- **Merge conflicts** — `mergeable_state: "dirty"`. The PR needs a rebase before merge.
- **Unused CI secrets** — `BASTION_SSH_KEY` and `KUBECONFIG_B64` are no longer referenced by the workflow. Consider removing them from repo secrets (outside this diff).

### Suppressed

- **APT_MIRROR sed duplication across 4 stages** — Dockerfiles don't support function extraction; the duplication is unavoidable and acceptable.
- **`docs/features/activity-and-ingest.md` not updated** — the local-dev INGEST_BASE_URL override is a dev-environment concern, not a change to the production telemetry contract. The doc covers production behavior; no update needed.
- **No epic CHANGELOG** — `docs/features/` has no CHANGELOG.md for AF-94 or a k3d/k8s epic. This appears to be a standalone infrastructure PR; the AGENTS.md epic CHANGELOG requirement doesn't apply.

---
*Reviewed by Hermes Agent*

compose.yml
```diff
@@ -113,7 +142,62 @@ services:
...
+    image: postgres:16
+    container_name: aai_litellm_db
+    profiles: ["k3d"]
+    restart: unless-stopped
+    environment:
+      POSTGRES_USER: litellm
+      POSTGRES_PASSWORD: litellm
+      POSTGRES_DB: litellm
+    volumes:
+      - aai_litellm_db_data:/var/lib/postgresql
```
aai-labs-agent: The main app's `db` service mounts at `/var/lib/postgresql/data` (repo convention), but litellm-db mounts at `/var/lib/postgresql` — missing the `/data` suffix. Both work (Postgres creates the data subdirectory inside the volume), but matching the existing pattern would be cleaner.

docker/k3d/k3d-up.sh
```diff
@@ -0,0 +1,188 @@
...
+printf "  export KUBECONFIG=%s\n" "${KUBECONFIG_HOST}"
+printf "  kubectl get pods -A\n\n"
+printf "  To run the API in Docker (make up) against this cluster, add to .env:\n"
+printf "    API_K8S_KUBECONFIG_PATH=/app/.k3d/kubeconfig-internal.yaml\n\n"
+printf "  Agents in k3d reach LiteLLM via:\n"
+printf "  http://host.docker.internal:%s\n\n" "${LITELLM_HOST_PORT}"
+printf "  Agents push telemetry back to the ingest API via:\n"
+printf "  http://host.docker.internal:%s/ingest/v1\n" "${INGEST_HOST_PORT}"
+printf "    make up         → served by the api container (port published)\n"
+printf "    make dev-api    → also run 'make dev-ingest' in a second shell\n\n"
```
aai-labs-agent: This summary message says `make dev-api → also run 'make dev-ingest' in a second shell`, but the `dev-api` Makefile target now starts ingest automatically alongside the main app (the trap-based co-start). This message is stale and contradicts both the Makefile and the README's "starts ingest alongside the main app" description.

aai-labs-agent:
## Hermes Agent Review — AF-94 Local k3d dev and CI environment

**Verdict: Comment** (1 warning, 1 suggestion — nothing blocking)

### Scope

13 files, +957/-22. Adds a local k3d cluster for developing the agent path, replaces the SSH-tunnel-based CI k8s integration with k3d, and fixes the telemetry wiring that silently dropped agent activity when the API ran outside the cluster.

### ⚠️ Warning

**`mergeable_state: dirty`** — this PR has conflicts with `staging`. A rebase is needed before merge. Given staging has had significant changes (including the 75+ migration chain), the conflicts may touch the CI workflow or test files.

### 💡 Suggestion

**LiteLLM port bound to all interfaces** (`compose.yml`) — `"${LITELLM_PORT:-7070}:4000"` publishes on `0.0.0.0`, not just loopback. The comment correctly explains this is needed for agent pods on native Linux to reach LiteLLM through the host. The proxy is gated by `LITELLM_MASTER_KEY`, so this is acceptable for local dev — just worth being aware that the proxy is reachable from the local network, not just localhost.

### AC / content review

- **k3d cluster setup** — `k3d-up.sh` is well-structured: checks dependencies, starts LiteLLM + Postgres, creates/adopts the cluster, writes dual kubeconfigs (host + in-container), seeds namespace + litellm secret, and adds CoreDNS `host.docker.internal` alias on native Linux. Good handling of the Docker Desktop vs native Linux difference.
- **Image loading** — `k3d-load-images.sh` has a useful `assert_tag_matches_version` check that catches env/config drift before it becomes an `ErrImagePull` at runtime.
- **CI workflow** — Replaces the SSH tunnel to `k8s.aai-labs.com` (which required `BASTION_SSH_KEY` and `KUBECONFIG_B64` secrets) with `AbsaOSS/k3d-action`. The `k3d-version` is pinned because the action's default 404s — good catch. This eliminates a class of CI failures (bastion connectivity, expired secrets) and removes the SSH tunnel attack surface.
- **Dockerfiles** — `APT_MIRROR` build arg defaults to `deb.debian.org`, so CI and existing builds are unchanged. The `sed` commands handle both `sources.list` and deb822 `.sources` layouts. Applied consistently to both `rust-build` and final stages in both Dockerfiles.
- **Telemetry wiring** — `INGEST_BASE_URL` override points pods at `host.docker.internal` instead of the in-cluster Service. The new test `test_start_agent_wires_telemetry_push_into_the_secret` verifies the secret carries the correct `INGEST_URL` and `INGEST_API_KEY`.
- **README** — Comprehensive setup guide with a `.env` table, troubleshooting section, and the APT_MIRROR escape hatch. The `cluster-down` vs `cluster-delete` distinction is well-explained.
- **`.env.spec`** — Adds all k3d-related values. Note: `LITELLM_MASTER_KEY` has no default and is required by the script, which is the right call (LiteLLM encrypts virtual keys with it, so it must be stable).

### Suppressed

- **No Helm chart bump** — PR body notes "No chart version bump: no `helm/` files changed." Correct per `operations.md`.
- **Hardcoded `litellm:litellm` DB credentials in compose** — local dev database, not production. Acceptable.

### Looks good

- `dev-api` now starts ingest alongside the main app, mirroring the container entrypoint — prevents the silent telemetry loss failure mode
- `extra_hosts: host.docker.internal:host-gateway` in compose handles the Linux case cleanly
- `.k3d/` directory mount is read-only and gitignored
- `cluster-down` semantics (stop, not destroy) match `db-down`/`redis-down` convention

---
*Reviewed by Hermes Agent*

sami-bre (changes requested):
Verified live on a from-scratch k3d cluster (Ubuntu, Docker 29.6): `cluster-up`, kubeconfig generation, namespace + `litellm` secret seeding, the CoreDNS host alias, image import, and the API driving the cluster from inside Docker.

**The core AC holds.** `KubernetesClient` lists/creates in k3d through `kubeconfig-internal.yaml` with no TLS SAN error, `LiteLLMClient.generate_key` mints and deletes a virtual key, and both pod→host hops (ingest, LiteLLM) return 200. Docs are unusually thorough.

## Blocking

**1. `compose.yml:170` — litellm-db persists to the wrong path (data loss).**
`aai_litellm_db_data:/var/lib/postgresql` mounts the *parent* of PGDATA. `postgres:16` declares `VOLUME /var/lib/postgresql/data`, so Docker creates an anonymous volume there that shadows the named one. `make clean` drops it and LiteLLM returns with an empty DB — every virtual key gone. That's the exact loss the "keep `LITELLM_MASTER_KEY` stable" check exists to prevent. The `db` service above uses the correct path.
```yaml
- aai_litellm_db_data:/var/lib/postgresql/data
```

**2. `.env.spec:20` — the placeholder admin password now works.**
`Replace_With_Secure_Password1` passes the policy, so `make setup && make up` silently bootstraps a platform admin with credentials committed to the repo, on an API published on all interfaces. The old non-compliant value failed startup, which forced a change. Keep it fail-closed.

**3. `docker/k3d/k3d-up.sh:106,109` — `\|` is GNU-only, breaks on macOS.**
Under `sed -E`, BSD/macOS reads `\|` as a literal pipe, not alternation, so both kubeconfig rewrites silently no-op and `kubeconfig-host.yaml` keeps `host.docker.internal`. Step 2's host path then can't reach the cluster on a Mac. Drop the backslashes: `(0\.0\.0\.0|host\.docker\.internal)`.

## Recommendations

- **Imported images don't survive `cluster-down` as claimed.** Hit this live — kubelet GC'd both base images under disk pressure (`usage=87 highThreshold=85`). They're unreferenced when no agent runs, and `IfNotPresent` against a private registry with no pull secret makes them unrecoverable → `ErrImagePull`. Consider `--kubelet-arg=image-gc-high-threshold=`, and add GC as a cause in the ErrImagePull pitfall.
- **`cluster-up` prints "ready" before CoreDNS is.** The README's own probe failed for me right after cluster-up and passed a minute later. `rollout restart` is `|| true`, so a real failure and "not ready yet" look identical.
- **No import-only path in `k3d-load-images.sh`.** It demands `GH_TOKEN` and tag==VERSION even when the image already exists locally — I couldn't use it and called `k3d image import` directly. A `k3d-import-images` target would cover this.
- **`k3d-load-images.sh:53-54`** requires both image vars even for single-target runs; `make k3d-load-hermes` fails without `OPENCLAW_IMAGE`. Move the checks inside the `TARGET` conditionals.
- **`k3d-load-images.sh:86`** reports "is not running" but only tests existence — a stopped cluster passes, then builds run for minutes before import fails.
- **`k3d-up.sh:196`** still prints `make dev-api → also run 'make dev-ingest' in a second shell`, which this PR made wrong.
- **`compose.yml:76`** `${API_K8S_KUBECONFIG_PATH-}` silently blanks a `.env`-provided `K8S_KUBECONFIG_PATH` (mine was set). `${API_K8S_KUBECONFIG_PATH:-${K8S_KUBECONFIG_PATH-}}` preserves it.
- **README host path** never sets `LITELLM_BASE_URL`; empty means key minting is skipped silently (`service.py:553`), so agents run but never answer.
- **CI lost RBAC coverage** — k3d gives the job cluster-admin, where the old kubeconfig authenticated as the restricted `agent-farm-user`. Applying that Role into k3d restores the check.
- **`test_agents.py:866`** leaks `INGEST_BASE_URL` into `os.environ` for the rest of the session; `set_env_variable` has no teardown.

kalibetre (approved):

PR #120, opened 2026-08-12
AF-256: add platform message and agent activity stats

AnaniyaT:
Adds two Platform View endpoints reporting cross-Organization message volume and Agent activity as bounded statistics, plus the Platform page surface that reads them.

## What's included

- **`api/domains/platform_admin/`** — new `routes.py`, `models.py`, `stats_service.py`. `GET /platform/stats/messages` and `GET /platform/stats/agents` behind `require_platform_admin()`, so no Active Organization is ever resolved. `PlatformStatsService` composes the conversations, agents and tool_calls **services**, never their repositories.
- **Reporting window** — a preset period or an explicit `from_date`/`to_date` range. Bucket size follows the span (minute under 2h, hour under 3d, day under 90d, week beyond) and can be pinned with `granularity`. A request needing more than 5,000 buckets is rejected, since `granularity` is caller-supplied.
- **Filters** — Organization, Agent, Agent creator, chat platform. Deliberately no sender filter: sender identity is outside the oversight boundary, so the contract cannot answer it.
- **Activity** — the union of chat messages and Tool Calls, deduplicated per bucket. Not a replay of `agent.started`/`agent.stopped`: those record administrative intent rather than liveness, emit nothing on ERROR transitions, and have no history before the outbox shipped. Tool Calls carry the weight here — the runtime plugins gate outbound *messages* on a user-triggered turn, so scheduled work is visible only through its tools.
- **`api/migrations/versions/c4d1e8f2a730_...`** — indexes `agent_chat_message (occurred_at, direction, agent_id)` and `tool_call (occurred_at, agent_id, organization_id)`. Every pre-existing index on both tables is `agent_id`-prefixed, so none can seek a platform-wide time range; the trailing columns keep the reads index-only.
- **`ui/src/features/platform-stats/`** — new feature module; `dashboard/platform/page.tsx` gains an Overview strip of current counts and an Activity section with controls, tiles and charts, below the existing administration links.
- **Unchanged** — no Organization-scoped route, DTO or `AuthorizationScope` path is touched. The existing agent-scoped conversations and tool-call reads keep going through `AgentAuthorization` exactly as before; the new aggregates are separate methods alongside them.

## Required repo config

None.

## Notes

- No chart version bump: no files under `helm/` changed (`docs/guidelines/operations.md:69`).
- Indexes are built without `CONCURRENTLY`, matching every other migration in this repo. That takes a `SHARE` lock and blocks INSERTs on the ingest write path for the length of the build (~0.75µs/row measured). The migration docstring records the trade-off.
- The aggregates all take an optional `organization_id`, so the later Organization dashboard can reuse them behind its own routes, read models and `AuthorizationScope` rather than widening these.
- Series buckets are emitted as UTC instants with an explicit offset — without one a browser parses them as local time and every chart label shifts by the viewer's timezone.
- Deliberately out of scope: deleting an Organization still cascades its message history away, so platform totals do not survive it. Agent-initiated ("cron") messages are dropped by both runtime plugins, so *Sent* means "replies to humans" until that is fixed upstream. Both warrant their own tickets.
- Per-Agent drill-down, Tool Call statistics, model usage and cost oversight remain with AF-250.

## Testing

- `make check-api`, `make check-migrations`, `make check-ui`, `make lint-ui` — all pass.
- `make test-ui` — 204 passed.
- `make test-api` — 1338 passed, including 24 new tests for this surface (authorization, window resolution, granularity selection, the bucket cap, UTC offsets, activity de-duplication across both telemetry streams, and the status partition).
- Run end to end against a local Postgres with the migration applied: verified both endpoints, every preset, the custom range, all four granularities, and the filters in the browser.
- Re-ran the full set after rebasing onto `staging`; that surfaced a second alembic head introduced by AF-251, which is why this migration is re-parented onto `a3d7f5e91c62`.


ui/src/features/platform-stats/utils.ts
```diff
@@ -0,0 +1,25 @@
+import { createQueryKeyStructure } from "@/shared/query-keys";
+
+import type { AgentPlatform, StatsPeriod } from "./schemas";
+
+export const platformStatsKey = createQueryKeyStructure("platform-stats");
+
+export const PERIOD_OPTIONS: { value: StatsPeriod; label: string }[] = [
```
aai-labs-agent: These three exports (`PERIOD_OPTIONS`, `DEFAULT_PERIOD`, `PERIOD_LABEL`) are unused in this PR — the panel imports only `PLATFORM_OPTIONS` from this module and uses its own preset system from `presets.ts`. The UI sends absolute `from_date`/`to_date` to the API, never the `period` parameter. If these are intended for a follow-up, I'd leave a comment; otherwise they're dead code worth removing.

api/domains/agents/repository.py
```diff
@@ -144,6 +146,166 @@ def count_agents_in_error(self) -> int:
...
+        deleted_at predicate cannot live in `_stats_predicates` — an Agent
+        deleted last week still existed in the buckets before.
+
+        `unit` is a postgres date_trunc unit ('hour' | 'day' | 'week'), chosen
+        from the window span so a one-day window is not a single bar and a
+        two-year one is not seven hundred. It also drives the generate_series
+        step, so the empty buckets are filled in at the same resolution.
+        """
+        predicates = self._stats_predicates(organization_id, agent_id, created_by_user_id, platform)
+        step = sa.text(f"interval '1 {unit}'")
```
aai-labs-agent: `sa.text(f"interval '1 {unit}'")` interpolates `unit` directly into SQL text — unparameterized. It's safe today because `unit` is always `window.granularity.value` (one of `minute`/`hour`/`day`/`week`), but the repository signature types it as `str`, wider than the enum. A future caller passing an unsanitized string would have SQL injection here. Constraining `unit` to `StatsGranularity` in the signature (or validating against a whitelist) would close the gap at the boundary.

api/domains/conversations/repository.py
```diff
@@ -61,6 +64,152 @@ def upsert_messages(self, messages: list[AgentChatMessage]) -> None:
...
+        the join is an inner join and deleted Agents still match, since nothing
+        here predicates on `Agent.deleted_at`.
+
+        occurred_at is timestamptz, so date_trunc would otherwise bucket in the
+        session TimeZone. The explicit UTC conversion makes the buckets
+        deterministic regardless of who is asking.
+        """
+        occurred_at_utc = sa.func.timezone("UTC", col(AgentChatMessage.occurred_at))
+        message_bucket = sa.func.date_trunc(unit, occurred_at_utc)
+        step = sa.text(f"interval '1 {unit}'")
```
aai-labs-agent: Same `sa.text(f"interval '1 {unit}'")` pattern as `agents/repository.py:231` — safe today via the enum, but the `unit: str` signature is wider than it needs to be. If the agents-side comment is addressed, the same fix should apply here for consistency.

aai-labs-agent (commented):
## Code Review Summary

**Verdict: Comment** — 2 suggestions, nothing blocking

Well-structured PR. The domain boundary is clean — Platform Admin routes compose other domains' services (never their repositories), authorization is enforced at the route via `require_platform_admin()`, no Active Organization is resolved, and the read models expose counts only. Migration head verified against `staging` (`a3d7f5e91c62` is a current head). Schema casing is handled by the shared API client's `transformKeys: true` default — backend snake_case is transformed to frontend camelCase before Zod validation, so the schemas match. `AgentPlatform` enum values match across backend and frontend. The `docs/features/platform-administration/CHANGELOG.md` entry for AF-256 is present and detailed, per the repo's review protocol for active epics.

### ✅ Looks Good

- **Authorization**: `require_platform_admin()` on both routes, rejects non-user sessions, no Active Organization resolved. Integration tests cover 401 (unauthenticated), 403 (regular user), and 200 (platform admin).
- **Activity deduplication**: union of message + tool-call agent IDs per bucket, with period-level distinct count. Unit + integration tests verify no double-counting across streams and across days.
- **Index design**: trailing `agent_id`/`organization_id` columns keep the platform-wide time-range reads index-only. Migration docstring documents the non-`CONCURRENTLY` trade-off.
- **Window resolution**: preset/custom range collapse to `[start, end)`, auto-granularity follows span thresholds, bucket cap at 5,000 rejects self-service DoS. All edge cases tested (naive datetimes, inverted ranges, oversized bucket counts).
- **Soft-deleted Agent handling**: historical message volume survives retirement (no `deleted_at` predicate on message joins). Agent inventory series reconstructs `existing`/`created` from `created_at`/`deleted_at` with a running total — O(buckets) instead of O(buckets × agents).
- **UI**: presets resolve to absolute instants client-side, `nowFloored` prevents query-key thrash, UTC offset on bucket labels tested, `atLeastAMinuteAfter` guards zero-width windows.
- **Test coverage**: 24 new backend tests (unit + integration) covering authorization, window resolution, granularity, bucket cap, UTC offsets, activity deduplication, filters, and the status partition.

### 💡 Suggestions

- **`ui/src/features/platform-stats/utils.ts`**: `PERIOD_OPTIONS`, `DEFAULT_PERIOD`, and `PERIOD_LABEL` are exported but unused — the UI uses its own preset system (`presets.ts`) and sends absolute dates, never the `period` query parameter.
- **`api/domains/agents/repository.py` & `conversations/repository.py`**: `sa.text(f"interval '1 {unit}'")` interpolates `unit` directly into SQL text. Safe today (always `StatsGranularity.value`), but the repository signature types `unit: str` — wider than the enum.

### Suppressed

- **Unused `context` parameter in route handlers**: `require_platform_admin()` is a dependency that enforces authorization; the `context` it returns isn't used in the handler body. Matches the repo pattern for auth-only dependencies — not dead code.
- **`dangerouslySetInnerHTML` in `chart.tsx`**: content is built from `ChartConfig` (CSS variable names and `var(--...)` values), not user input. No XSS surface.
- **`AgentPlatform` enum parity**: backend `["slack", "teams", "telegram"]` matches frontend `z.enum(["slack", "teams", "telegram"])`. ✅

---
*Reviewed by Hermes Agent*

PR #121, opened 2026-08-12
Af 248 replace mailgun with cloudflare services

dominykas-aai-labs:


api/infrastructure/email/client.py
```diff
@@ -23,98 +25,192 @@
...
+# strip inline <svg>, so templates reference `cid:LOGO_CONTENT_ID` and we ship the
+# rasterised PNG as an inline attachment whenever a template asks for it.
 LOGO_CONTENT_ID = "agent-barn-logo"
 _LOGO_PATH = os.path.join(os.path.dirname(__file__), "assets", "logo.png")
 
+_SEND_URL = "https://api.cloudflare.com/client/v4/accounts/{account_id}/email/sending/send"
+
+# Split timeouts rather than one flat budget: sync FastAPI routes run in a threadpool, so a
+# slow Cloudflare must not pin a worker for the whole window.
+_TIMEOUT = httpx.Timeout(connect=5.0, read=10.0, write=10.0, pool=5.0)
```
AnaniyaT: the timeouts are per-attempt, not a total budget — 3 × (5s connect + 10s read) + backoff is ~47s worst case. invite and forgot-password send in-request (`auth/service.py:293`, `:315`), and sync routes share one 40-thread anyio pool, so a cloudflare stall parks threads that every other endpoint needs. needs an overall deadline across attempts.

api/infrastructure/email/service.py
```diff
@@ -200,32 +201,38 @@ def send_agent_lifecycle_email(
...
-                subject=f"Agent {agent_name} {lifecycle_action}",
-                receiver_name=receiver_name,
-                receiver_email=receiver_email,
-                attributes=[
-                    EmailTemplateAttribute(name="user_name", value=receiver_name or receiver_email),
-                    EmailTemplateAttribute(name="agent_name", value=agent_name),
-                    EmailTemplateAttribute(name="lifecycle_action", value=lifecycle_action),
-                ],
-            )
             email = self.create_email(email_template)
```
AnaniyaT, line 229: only `mjml_to_html` is inside the terminal-error wrapper — `read_template` and the jinja `Template(...)`/`render()` calls sit outside it. so a renamed or syntactically broken template escapes as a plain exception, hits the catch-all in `event_handlers.py:69`, and gets marked retryable. that burns the full dramatiq budget (20 attempts, ~4h) on a failure that can never succeed, instead of dead-lettering immediately.

api/tests/conftest.py
```diff
@@ -65,3 +70,19 @@ def clear_openrouter_models_cache():
...
+    invite reaches the real EmailClient. Block the transport for the whole suite: a test that
+    genuinely exercises sending patches this target itself, which overrides this fixture."""
+
+    def _refuse(*args, **kwargs):
+        raise RuntimeError(
+            "Outbound email attempted in tests. Patch api.infrastructure.email.client.httpx.post "
+            "or bind MockEmailModule if this send is intentional."
+        )
+
+    with patch("api.infrastructure.email.client.httpx.post", side_effect=_refuse):
```
AnaniyaT: patching `api.infrastructure.email.client.httpx.post` resolves through the httpx module object, so it replaces `httpx.post` process-wide — litellm, slack, gmail and the google oauth routes all get the email error if they're ever unmocked (`litellm/client.py:47` only catches `httpx.HTTPError`, so it'd escape as a bare `RuntimeError`). we already have `MockEmailModule` and `prepare_injector` composes defaults + overrides, so bind it there instead. if you want a global guard too, patch `EmailClient.send`, not the httpx module.

PR #122, opened 2026-08-13
Tell agents what each integration can do, not just its --profile

sami-bre:
## Summary

Follow-up to #109. Slack channel data access works, but agents didn't reliably know they had it — an agent with the Slack skill attached told a user *"I can't search channel history, browse past messages, or look up attachments. This is a limitation of how I'm set up in Slack"*, then answered the same question correctly four minutes later with a full file listing. Same agent, same channel, opposite answers.

Neither failure was a credential or capability gap. Both are gaps in the context that is **always** loaded, and both are provider-agnostic — Slack just surfaced them first, because it's the one integration that collides with the agent's own transport.

## What changed

**`build_integrations_policy_md` (AGENTS.md) — state the capability, not just the slug.**

A Slack-only agent's entire integrations footprint was:

```
- **Slack**: `--profile slack-work`
```

Nothing there connects *"are there files attached in this channel?"* to `slack-work`. With no token in context matching the question, the model falls back on its prior about what a Slack bot can do. Each line now closes with a one-clause summary of the command surface:

```
- **Slack**: `--profile slack-work` — read channel data: list channels, list and
  download files and attachments, read bookmarks, links, canvases (read-only)
- **GitHub**: `--profile github-work` → aai-labs/agent-farm — PRs (diff, files,
  reviews, comments), issues, branches, repo source, Actions runs
```

Each blurb is taken from the commands actually documented in that provider's `aai_cli_skills/<provider>.py`, not written from memory. The calendars ship no aai-cli skill doc, so they get no blurb rather than an invented one, and render exactly as before.

**`_TOOL_CONTEXT_PROVIDERS` (TOOLS.md) — stop skipping most providers.**

The "Configured Integrations" block — the one that says *credentials are already in place, do not ask the user to re-provide them* — was gated on a hardcoded `{GITHUB, JIRA, CONFLUENCE, BITBUCKET}`. Slack, Gmail, Zoho Mail, Pipedrive and the calendars produced **no block at all**, which is why both failure transcripts end with the agent asking the user to share the file directly.

It's now keyed off `PROFILE_SLUGS`, so any provider reachable through a `--profile` is covered the moment it gets one. Providers with no site/repo metadata worth printing render as a plain `- **Slack** (`slack-work`)` line — "configured" is the whole point. Firecrawl has no profile and is still correctly excluded.

## Tests

Three tests asserted the old omissions (`test_tool_context_md_omits_slack`, `..._empty_when_only_pipedrive`, `..._omits_non_aai_cli_providers`) and are updated to the new behaviour with the reasoning recorded. Four new tests cover the capability blurbs, including the repo-scoped GitHub/Bitbucket branch (built by a different code path) and the deliberate no-blurb case.

`check-api`, `check-ui`, `check-migrations` clean. Full unit suite: 655 passed. Integration tests untouched by these strings — no test outside `test_aai_cli_artifacts.py` references either block.

## What this does *not* fix

The underlying discovery problem, which is why the behaviour was intermittent rather than simply absent.

Skill docs are written to `<workspace>/skills/aai-cli/<service>_skill.md`. On OpenClaw that is the **highest-precedence skill root** — but discovery matches the exact basename `SKILL.md` (`src/skills/runtime/refresh.ts:423`), so our files never enter the `<available_skills>` block that OpenClaw's own system prompt instructs the agent to scan. Hermes indexes `~/.hermes/skills/<name>/SKILL.md`; we write to `/workspace/skills/`, missing on both path and filename.

So on both runtimes the *only* pointer to these docs is prose the agent may or may not act on — which is exactly the coin-flip behaviour observed. Converting the seeded docs to real `SKILL.md` files with `name`/`description` frontmatter is the structural fix and wants its own PR.

### DIFF
--- a/api/domains/agents/aai_cli_artifacts.py
+++ b/api/domains/agents/aai_cli_artifacts.py
@@ -264,12 +264,13 @@ _PROFILE_BUILDERS: dict[SecretProvider, Callable[..., str]] = {
 }
 
 
-_TOOL_CONTEXT_PROVIDERS = {
-    SecretProvider.GITHUB,
-    SecretProvider.JIRA,
-    SecretProvider.CONFLUENCE,
-    SecretProvider.BITBUCKET,
-}
+# Every provider reachable through an aai-cli --profile gets a "Configured Integrations"
+# line. This was previously limited to the four repo/issue trackers, which left Slack,
+# Gmail, Zoho Mail, Pipedrive, and the calendars with no "credentials are already in
+# place" note at all — so those agents would tell the user they had no access, or ask for
+# a token that was already mounted. Keyed off PROFILE_SLUGS so a new provider is covered
+# the moment it gets a profile.
+_TOOL_CONTEXT_PROVIDERS = frozenset(PROFILE_SLUGS)
 
 
 def build_tool_context_md(decrypted: Mapping[SecretProvider, SecretContent]) -> str:
@@ -291,7 +292,7 @@ def build_tool_context_md(decrypted: Mapping[SecretProvider, SecretContent]) -> 
     ]
     for provider in SecretProvider:
         content = decrypted.get(provider)
-        if content is None:
+        if content is None or provider not in PROFILE_SLUGS:
             continue
         if isinstance(content, GithubContent):
             base = PROFILE_SLUGS[SecretProvider.GITHUB]
@@ -323,6 +324,11 @@ def build_tool_context_md(decrypted: Mapping[SecretProvider, SecretContent]) -> 
                     f"- **Bitbucket** (`{base}`): workspace `{content.workspace}` "
                     f"({content.email}) — no repository configured; pass --repo explicitly"
                 )
+        else:
+            # Providers with no site/repo metadata worth printing still belong here: the
+            # point of this block is "credentials are already in place", which is exactly
+            # what a Slack- or Gmail-only agent was missing.
+            lines.append(f"- **{_INTEGRATION_LABELS[provider]}** (`{PROFILE_SLUGS[provider]}`)")
     return "\n".join(lines) + "\n"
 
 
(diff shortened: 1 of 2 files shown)


aai-labs-agent:
## Code Review — PR #122

**Verdict: Looks good — no blocking issues.**

### What this PR does

Two focused changes to agent context generation in `aai_cli_artifacts.py`:

1. **`build_tool_context_md` (TOOLS.md block)** — `_TOOL_CONTEXT_PROVIDERS` expands from a hardcoded set of 4 repo/issue-tracker providers to `frozenset(PROFILE_SLUGS)`, so every provider with an aai-cli profile now gets a "credentials are already in place" line. An `else` branch renders providers without site/repo metadata as a plain `- **Label** (`slug`)` entry.

2. **`build_integrations_policy_md` (AGENTS.md block)** — Adds `_INTEGRATION_CAPABILITIES`, a one-clause description of each provider's command surface, appended to its integrations line. Calendars (no aai-cli skill doc) are deliberately omitted and render as before.

### ✅ Looks Good

- **Well-motivated.** The PR body traces both failure modes (agent denying access it had, then answering correctly minutes later) to a concrete context gap, and the fix targets exactly that gap.
- **Sourced from real data.** Capability blurbs are taken from the command surface in `aai_cli_skills/<provider>.py`, not invented. The deliberate omission of calendars (no skill doc) is the right call.
- **Clean refactor in `build_integrations_policy_md`.** The transition from direct `lines.append()` to building `line` then appending with capability is minimal and readable. The repo-scoped branch (GitHub/Bitbucket) is covered by the same capability append.
- **Test coverage is thorough.** Three old omission-asserting tests are updated with reasoning recorded in comments. Four new tests cover: Slack capability blurb, repo-scoped GitHub line with capability, calendar no-blurb case, and the multi-repo path. The Firecrawl-excluded test is kept with an updated comment.
- **No unrelated churn.** 2 files, 88 additions, 22 deletions — all on target.
- **No helm changes** — no chart version bump needed per `docs/guidelines/operations.md`.

### Notes

- **`_INTEGRATION_LABELS[provider]` in the new `else` branch (line 331)** uses `[]` indexing rather than `.get()`. If a future provider is added to `PROFILE_SLUGS` without a matching `_INTEGRATION_LABELS` entry, this would `KeyError` at runtime. Not a current bug — all 11 providers in `PROFILE_SLUGS` are covered — and the same pattern already exists in `build_integrations_policy_md`. Mentioning only as a forward-looking note; no action needed in this PR.
- **`docs/features/integrations.md`** documents the materialization flow at a high level ("appends tool/integration policy to rendered template content") and points to `aai_cli_artifacts.py` as the authoritative source. No doc update required — the change enhances existing context generation without altering an invariant, boundary, or state model.
- **Epic CHANGELOG check**: `docs/features/integrations.md` is a single file, not a directory with a CHANGELOG — the multi-PR epic rule doesn't apply.

---
*Reviewed by Hermes Agent*

PR #127, opened 2026-08-15
AF-261 Skills v2 draft/publish authoring, built-in forking, per-version deletion, and agent version pinning

kalibetre:
## Summary

Skills v2 reworks the skill authoring and version management lifecycle so organizations can safely edit, version, recover from, and customize skills without disrupting running agents.

Replaces the old upload-and-replace ZIP model with a draft/publish flow, adds built-in skill forking, introduces per-version deletion with agent-pin-aware protections, and pins each agent skill assignment to an explicit version so publishing a newer version never silently changes what a running agent mounts.

Closes [AF-261](https://aai-labs.atlassian.net/browse/AF-261)

## What changed

**Draft/publish authoring** — Content edits go through a single in-flight draft per skill. A manager saves and reviews the draft, then publishes it as a new immutable version. The skill detail page shows file cards, the draft editor, and a version history panel.

**Built-in skill forking** — An organization can fork a built-in skill into an editable, org-scoped copy seeded from the built-in's current content. The fork is a one-time snapshot with its own lineage; the built-in stays read-only.

**Per-version deletion** — A manager can prune individual version snapshots instead of deleting an entire lineage. The last remaining version and any version pinned by an agent are always protected. Replaces whole-lineage delete.

**Agent skill version pinning** — Each agent skill assignment pins an exact version at assign/update time. Publishing a newer version never moves an existing pin. Recovery from a bad version is per-agent: re-pin instead of lineage-level restore.

**Audit fixes** — Tenant-scoped skill lookups, atomic pin validation (no partial state on invalid pins), legacy migration path validation, composite FK for delete/pin race safety, draft cache invalidation, draft hydration on reload.

[AF-261]: https://aai-labs.atlassian.net/browse/AF-261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


api/migrations/versions/04088157c78c_merge_heads_after_rebasing_skills_v2_.py
```diff
@@ -0,0 +1,23 @@
...
+Revises: a1b2c3d4e6f7, c4d1e8f2a730
+Create Date: 2026-08-17 12:29:22.999052
+
+"""
+
+from collections.abc import Sequence
+
+# revision identifiers, used by Alembic.
+revision: str = "04088157c78c"
+down_revision: str | Sequence[str] | None = ("a1b2c3d4e6f7", "c4d1e8f2a730")
```
aai-labs-agent: 🔴 **Critical — 2 migration heads after merge (deploy blocker)**

`c4d1e8f2a730` is **not a head on staging**. Staging's chain continues past it:

```
c4d1e8f2a730 → d5e2a9f4c781 (merge_discord_and_staging_heads)
```

So this merge joins a revision that's already consumed. After the PR merges, Alembic will see two heads:

- `04088157c78c` (this migration — nothing chains from it)
- `d5e2a9f4c781` (staging's head — nothing in this PR chains from it)

Deploy will fail with *"Multiple heads are present; please specify which head to begin from."*

**Fix:** change `down_revision` to reference `d5e2a9f4c781` (the actual staging head) instead of `c4d1e8f2a730`, or add one more merge migration joining `04088157c78c` with `d5e2a9f4c781`.

I verified this by batch-fetching all 75 staging migration files, extracting their `revision`/`down_revision`, and computing heads. `c4d1e8f2a730` is consumed by `d5e2a9f4c781` — confirmed.

api/domains/skills/repository.py
```diff
@@ -21,6 +30,285 @@ class SkillRepository:
...
+        """Files of each skill's *pinned* version, keyed by skill id.
+
+        Used when materializing an agent's skills at start, so mounting resolves
+        the exact version an agent pinned rather than the lineage's latest.
+        """
+        if not pins:
+            return {}
+        with Session(self.delegate.engine) as session:
+            version_rows = []
+            for pin in pins:
```
aai-labs-agent: 💡 **Suggestion — N+1 query loop**

This loop fires one `SELECT skill_version WHERE skill_id = ? AND version = ?` per pin. For an agent with N skills, that's N queries.

Could be batched into a single query:
```python
conditions = []
for pin in pins:
    conditions.append(
        sa.and_(col(SkillVersion.skill_id) == pin.skill_id,
                col(SkillVersion.version) == pin.version))
rows = session.exec(select(SkillVersion).where(sa.or_(*conditions))).all()
```

Or a tuple-IN: `WHERE (skill_id, version) IN (...)`. Impact is low in practice (agents have a handful of skills), but this is the one method in the hot path that doesn't batch — every other batch method in this repo does (`get_latest_version_numbers`, `get_draft_skill_ids`).

api/domains/skills/repository.py
```diff
@@ -21,6 +30,285 @@ class SkillRepository:
...
+        self,
+        draft_id: UUID,
+        files: list[tuple[str, str]],
+        *,
+        description: str | None = None,
+        required_providers: list[str] | None = None,
+    ) -> SkillDraft:
+        with Session(self.delegate.engine) as session:
+            draft = session.get(SkillDraft, draft_id)
+            assert draft is not None
```
aai-labs-agent: ⚠️ `assert` is stripped under `python -O`

If `draft` is somehow `None` (e.g., a race between the service's existence check and this call), `assert` is a no-op under `-O` and the next line (`draft.description = ...`) crashes with an opaque `AttributeError`. No other repository in the repo uses `assert` for null guards.

Suggested fix:
```python
draft = session.get(SkillDraft, draft_id)
if draft is None:
    raise ValueError(f"Draft {draft_id} not found")
```

api/domains/skills/service.py
```diff
@@ -142,65 +181,236 @@ def update_skill(self, skill_id: UUID, data: SkillUpdate, context: CurrentUserCo
...
-            self._validate_zip(updated["zip_content"])
-            skill.zip_content = updated["zip_content"]
+        # The slug (and therefore the mount directory) is deliberately not
+        # recomputed on rename: renaming a skill must not move its files or
+        # invalidate paths referenced from inside its own markdown.
         if "name" in updated:
             skill.name = updated["name"]
-            skill.tools_pointer = f'You can use "{skill.name}" skill in the ./skills folder'
+        if "description" in updated:
+            skill.description = updated["description"]
```
aai-labs-agent: ⚠️ **Docs-vs-code mismatch: `PATCH /{skill_id}` allows description/required_providers but docs say name-only**

`docs/features/templates-and-skills.md` says:
> *"`PATCH /{skill_id}` is metadata-only (name)"*
> *"Description and required providers are staged on the draft and only applied on publish, so the published version's metadata stays frozen until a draft is published."*

But `SkillUpdate` includes `description` and `required_providers`, and this method applies them directly to the skill row — bypassing the draft flow.

This creates a silent overwrite: if a manager updates description via `PATCH /{skill_id}` while a draft is in flight, `publish_draft` will overwrite the direct update with the draft's staged value (the draft was seeded from the pre-PATCH description).

Either narrow `SkillUpdate` to name-only (matching the docs), or update the docs to document that `PATCH /{skill_id}` also accepts description and required_providers and explain the interaction with draft-staged metadata.

aai-labs-agent (changes requested):
## Code Review Summary

**Verdict: Changes Requested** — 1 critical (deploy blocker), 2 warnings, 1 suggestion.

Reviewed the full diff (62 files, +6022/-1154) against the AF-261 acceptance criteria and the repo's `AGENTS.md` review protocol. The design is well-structured — the draft/publish flow, version pinning, forking, and per-version deletion are all cleanly implemented with proper authorization, tenant scoping, and race-condition protections. The migration chain is the blocker.

### 🔴 Critical

- **Migration chain has 2 heads after merge** — `api/migrations/versions/04088157c78c_merge_heads_after_rebasing_skills_v2_.py` merges with `c4d1e8f2a730`, but that revision is **not a head on staging**. Staging's chain continues `c4d1e8f2a730 → d5e2a9f4c781` (merge_discord_and_staging_heads). After this PR merges, Alembic will report two heads (`04088157c78c` and `d5e2a9f4c781`) and fail at deploy with *"Multiple heads are present."* The fix: change the final merge migration's `down_revision` to reference `d5e2a9f4c781` (the actual staging head) instead of `c4d1e8f2a730`, or add one more merge migration joining `04088157c78c` with `d5e2a9f4c781`.

### ⚠️ Warnings

- **`PATCH /{skill_id}` allows description/required_providers despite docs saying name-only** — `docs/features/templates-and-skills.md` says *"`PATCH /{skill_id}` is metadata-only (name)"* and *"Description and required providers are staged on the draft and only applied on publish, so the published version's metadata stays frozen until a draft is published."* But `SkillUpdate` includes `description` and `required_providers`, and `update_skill` applies them directly to the skill row. A direct PATCH can be silently overwritten by a subsequent draft publish (since `publish_draft` copies the draft's staged metadata onto the skill row). Either narrow `SkillUpdate` to name-only, or update the docs to document the interaction.

- **`assert` in production repository code** — `repository.py:243` uses `assert draft is not None` in `update_draft_files`. Under `python -O`, `assert` is stripped and the code would crash on the next line (`draft.description = ...`) with an opaque `AttributeError` instead of a clear error. No other repository in the repo uses `assert` for null guards. Replace with an explicit `if draft is None: raise ...` check.

### 💡 Suggestions

- **N+1 query in `get_files_for_skill_versions`** — `repository.py:111` loops over `pins` with a separate `SELECT` per pin inside the loop. For an agent with N skills, this is N queries. Could be batched into a single `SELECT skill_file JOIN skill_version WHERE (skill_id, version) IN (...)`. Impact is low in practice (agents typically have a handful of skills), but it's the one method in the hot path that doesn't batch.

### ✅ Looks Good

- **Tenant scoping** — `get_by_id_for_org` correctly returns 404 (not 403) for other orgs' skills, scoped by `organization_id == org_id OR organization_id IS NULL`. Draft lookups go through the same scoped path.
- **Authorization** — All skill service methods check `SKILL_MANAGE` or `SKILL_READ`. Built-in skills reject modification/forking-with-non-aai-source with 403. Pin validation (`_resolve_skill_pins`) runs before any persistence — no partial state on invalid pins.
- **Race condition protection** — `delete_skill_version` checks `is_skill_version_pinned` at the service level and catches `IntegrityError` from the composite FK (`fk_agent_skill_pinned_version`, `ondelete="RESTRICT"`) as a safety net for concurrent delete/pin races.
- **File path validation** — `files.py` `normalize_path`/`validate_files` is thorough: traversal, absolute paths, junk metadata, case-insensitive duplicates, per-file/total size caps, file-count limits. The migration backfill (`_readable_entries`) inlines the same contract so legacy archives are validated identically to the API boundary.
- **Agent start** — Logs a warning when a pinned version's files are missing; reports file collisions between skills sharing a `root_dir`.
- **Soft-delete exclusion** — `get_skills_for_agents_with_versions` joins with `Agent.deleted_at IS NULL`, so batch loading can't leak skills from soft-deleted agents.
- **Documentation** — Feature docs (`templates-and-skills.md`, `agents.md`), two ADRs (pin-skill-versions, per-version deletion) are all updated in the same diff. The seeder now publishes versions only when content changes, avoiding version inflation on restart.

### Suppressed

- **Transaction boundary in `create_skill`/`fork_skill`** — `save(skill)` then `publish_version(...)` then `save_new_draft(...)` are separate transactions (each opens its own Session). If `publish_version` fails, the skill row is orphaned with no version. Checked the repo convention: `self.delegate.save()` opens its own Session everywhere, so this is the established pattern, not a regression. Not flagging because it matches the existing architecture, but worth noting that a future refactor to wrap multi-step persistence in one session would close the gap.
- **`publish_draft` converting `list[str]` → `list[SecretProvider]`** — If a draft's `required_providers` somehow contained an invalid string, `SecretProvider(p)` would raise `ValueError`. But the API boundary validates via `SkillDraftUpdate.required_providers: list[SecretProvider]`, so invalid values can't enter through the normal flow. Not a finding.
- **Cross-org `is_pinned_by_agent` visibility** — `SkillVersionRead.is_pinned_by_agent` is org-scoped, so a version pinned by another org's agent wouldn't show a badge but would still be protected by the FK constraint. This only matters for built-in skills (globally visible), and built-ins reject version deletion with 403, so the scenario is unreachable. Not a finding.

### Acceptance Criteria Cross-Check

| 1 | Create skill, draft editor, save/discard/publish, one in-flight draft | ✅ | `create_skill`, `start_skill_draft`, `update_skill_draft`, `discard_skill_draft`, `publish_skill_draft`; `uq_skill_draft_skill_id` unique constraint |
| 2 | Member can read but not manage drafts; built-ins reject draft ops | ✅ | `_require_draftable_skill` checks `SKILL_MANAGE`, rejects `AAI_CLI` with 403 |
| 3 | Reload hydrates from persisted draft, not published version | ✅ | `get_skill_draft` reads from `skill_draft`/`skill_draft_file` |
| 4 | Fork built-in → org-scoped custom, carries metadata, opens draft | ✅ | `fork_skill` + `POST /{skill_id}/fork`; built-in lineage untouched |
| 5 | Fork name preserved/disambiguated; forking custom/other-org rejected | ✅ | `_allocate_fork_name`; source check `if source.source != SkillSource.AAI_CLI: raise 400` |
| 6 | Per-version delete; last version & pinned protected; UI shows pins | ✅ | `delete_skill_version`; `len(versions) <= 1` → 409; `is_skill_version_pinned` → 409; `SkillVersionRead.is_pinned_by_agent` |
| 7 | Deleting unpinned version doesn't affect agents; whole-lineage delete removed | ✅ | Old `DELETE /{skill_id}` removed; built-ins rejected with 403 |
| 8 | Agent pins exact version at assign/update; publish doesn't move pin | ✅ | `agent_skill.pinned_version`; `publish_version`/`publish_draft` never update `AgentSkill` rows; `re_pin_skill` |
| 9 | Pins submitted with create/update; omitted = latest; invalid rejected before persistence | ✅ | `_resolve_skill_pins` preflight before `save`/`save_skills` in both `create_agent` and `update_agent` |
| 10 | UI version selector; removing skill after version change doesn't cause 500; description clearing works | ✅ | `agent-skills-tab.tsx` version selector; `update_agent` applies `removed_skill_ids` before pin re-application |
| 11 | Draft ops tenant-scoped at repo level; batch loading excludes soft-deleted | ✅ | `get_by_id_for_org` scoped; `get_skills_for_agents_with_versions` joins `Agent.deleted_at IS NULL` |
| 12 | DB constraint prevents deleting pinned version under race; start logs missing files | ✅ | Composite FK `fk_agent_skill_pinned_version` `ondelete="RESTRICT"`; `IntegrityError` catch; warning log in `start_agent` |
| 13 | Legacy backfill validates paths/sizes/counts; invalid skipped & reported | ✅ | Migration `_readable_entries` + `_normalize_path` with full validation; `logger.warning` for skipped/empty |
| 14 | Feature docs + ADRs document all new capabilities | ✅ | `templates-and-skills.md`, `agents.md`, two ADRs — all in diff |

---
*Reviewed by Hermes Agent*

aai-labs-agent:
## Additional Findings — from comparison with @Dominykas's review

After posting the initial review, I compared findings with a colleague's review. The items below are from that review — each independently verified against the code at this commit. They are posted as inline comments alongside the original findings.

**One critical issue from my initial review remains unaddressed:** the migration chain has 2 heads after merge (deploy blocker) — see the inline comment on `04088157c78c_merge_heads_after_rebasing_skills_v2_.py:13`.

### 🔴 Blocking (from colleague's review, verified)

1. **Name field in draft editor silently discards input** — `handleSaveDraft` sends files/description/providers to `PATCH /draft` but not `name`. `useUpdateSkill` (the only mutation that persists name) is exported but imported nowhere. The old `skill-drawer.tsx` edit form was deleted, so there is currently no way to rename a skill from the UI. The e2e test `settings-skill-detail-page.spec.ts:133` installs `interceptUpdateSkillRequest()` but never asserts it fired.

2. **Draft metadata cannot be cleared via `null`** — `update_draft_files` (repository.py:249) uses `if description is not None:` as a guard, so `None` means "skip" rather than "clear". `SkillDraftUpdate.description` defaults to `None`, so omitting or sending `null` silently keeps the previous value. The AC passes only because the UI sends `""`. `update_skill` already does this correctly with `exclude_unset`.

3. **No API test coverage for draft metadata staging** — `grep -i description api/tests/integration/test_skills.py` returns nothing. The draft metadata staging + publish-apply logic is untested at the API boundary, including the explicitly-called-out clearing behaviour.

### ⚠️ High Risk (from colleague's review, verified)

4. **`ON DELETE RESTRICT` on composite FK can break organization deletion** — PostgreSQL evaluates `RESTRICT` immediately during cascade processing; if the `skill_version` cascade fires before the `agent_skill` cascade (order depends on RI trigger OID sort — not guaranteed), the whole org delete aborts with a FK violation. `delete_organization` only blocks on active agents; soft-deleted agents keep their `agent_skill` rows. Fix: switch to `NO ACTION` (defers check to end-of-statement) — same protection for direct version delete, but sibling cascades can satisfy it.

5. **Cross-org version-existence oracle in `_resolve_skill_pins`** — pin preflight runs before the skill accessibility check in both `create_agent` (line 710) and `update_agent` (line 1587). `get_version(skill_id, pin.version)` is not org-scoped, so for a known skill UUID from another org, the error message distinguishes "version doesn't exist" from "version exists" — both 404, but the detail differs. The AC explicitly requires tenant scoping at the repository level.

### Medium (from colleague's review, verified)

6. N+1 in `get_files_for_skill_versions` — *(already posted as inline comment in initial review)*

7. **Skill in both `skill_ids` and `removed_skill_ids` is silently dropped** — removals apply first, then `re_pin_skill` returns silently when the row is gone. Previous code re-added via `add_skill`. Not reachable from the current UI, but a silent API behaviour change.

8. **`_allocate_slug` / `_allocate_fork_name` are read-then-write races** — two concurrent creates in one org pick the same slug; the loser hits `uq_skill_organization_slug` as an unhandled `IntegrityError` → 500.

9. `fork_skill`/`create_skill` not atomic — *(already noted as suppressed in initial review)*

10. **`GET /{skill_id}/draft` doesn't reject built-ins** — uses `_get_or_404` + `SKILL_MANAGE` instead of `_require_draftable_skill`, so a built-in returns 404 where every other draft operation returns 403. AC says all draft operations are "rejected for built-ins."

### Low / cleanup (from colleague's review, verified)

11. Dead code: `SkillRepository.delete()`, `get_skills_for_agents()` (superseded), `useUpdateSkill()`, `SkillDetailRead.is_assigned_to_agent`/`isAssignedToAgent`.

12. **"v0" flash** — `Math.max(0, ...)` renders `v0` while `useSkillVersions` is pending. `detail.version` is already loaded and correct — use it as fallback.

13. **Spurious dirty state** — `hasPendingChanges` counts any entry in `pendingPins`, so re-selecting the already-pinned version enables Apply. Compare against `skill.version` before marking dirty.

14. Doc drift on `PATCH /{skill_id}` — *(already posted as inline comment in initial review)*

---
*Findings 1-5, 7-8, 10-13 surfaced during comparison with @Dominykas's review. Each was independently verified against the code at commit `1cbbd50a`.*

ui/src/features/skills/components/skill-detail-page.tsx
```diff
@@ -0,0 +1,603 @@
...
+  }
+
+  async function handleSaveDraft() {
+    if (!localFiles || !validateEntry(localFiles)) return;
+    setFileError(null);
+    try {
+      const draft = await updateDraft.mutateAsync({
+        skillId,
+        files: localFiles,
+        description: description.trim(),
```
aai-labs-agent: 🔴 **Blocking — Name field silently discarded (from @Dominykas's review, verified)**

`handleSaveDraft` sends `files`, `description`, `requiredProviders` to `PATCH /draft` but **not `name`**. `useUpdateSkill` — the only mutation that persists a name — is exported from `use-skill-mutations.ts:69` but imported nowhere. Since `skill-drawer.tsx` (which held the old edit form) was deleted, there is currently **no way to rename a skill from the UI**.

The user renames, clicks Save draft, gets a success toast, and the rename is gone on reload.

The e2e test `settings-skill-detail-page.spec.ts:133` installs `interceptUpdateSkillRequest()` but never asserts it fired — stale scaffolding.

*Fix:* wire `useUpdateSkill` into `handleSaveDraft` (fire when `name !== detail.name`), or make the name read-only in the editor and drop `useUpdateSkill`. Then make the e2e test assert whichever contract is chosen.

api/domains/skills/repository.py
```diff
@@ -21,6 +30,285 @@ class SkillRepository:
...
+            session.exec(  # type: ignore[call-overload]
+                delete(SkillDraftFile).where(col(SkillDraftFile.skill_draft_id) == draft_id)
+            )
+            for path, content in files:
+                session.add(SkillDraftFile(skill_draft_id=draft_id, path=path, content=content))
+            if description is not None:
```
aai-labs-agent: 🔴 **Blocking — Draft metadata cannot be cleared via `null` (from @Dominykas's review, verified)**

```python
if description is not None:
    draft.description = description
if required_providers is not None:
    draft.required_providers = required_providers
```

`SkillDraftUpdate.description` defaults to `None`, so a client that omits `description` or sends `null` silently keeps the previous value. The AC "clearing a skill's description in the editor clears it server-side" passes only because the UI happens to send `""` (`description.trim()` at `skill-detail-page.tsx:218`) — and it persists `""` rather than `NULL`.

`update_skill` (service.py:183) already does this correctly with `model_dump(exclude_unset=True)`. Apply the same pattern here: an explicit `null` clears, an omitted field is untouched. Keep `required_providers` consistent.

api/migrations/versions/5f6e7d8c9b0a_add_pinned_version_to_agent_skill.py
```diff
@@ -0,0 +1,55 @@
...
+        "skill_version",
+        ["skill_id", "pinned_version"],
+        ["skill_id", "version"],
+        ondelete="RESTRICT",
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint("fk_agent_skill_pinned_version", "agent_skill", type_="foreignkey")
+    op.drop_column("agent_skill", "pinned_version")
```
aai-labs-agent, line 55: ⚠️ **High — `RESTRICT` can break organization deletion (from @Dominykas's review, verified)**

`ondelete="RESTRICT"` evaluates immediately during cascade processing. When deleting an org:

1. `organization → skill` cascade fires
2. Deleting a `skill` row fans out to `skill_version` (CASCADE) and `agent_skill` (CASCADE)
3. If the `skill_version` cascade fires before the `agent_skill` cascade (order = RI trigger OID sort, not guaranteed), the `RESTRICT` check on `agent_skill.(skill_id, pinned_version) → skill_version.(skill_id, version)` fails because `agent_skill` rows still reference the `skill_version` rows being deleted.

This is reachable: `delete_organization` only blocks on _active_ agents, and soft-deleted agents keep their `agent_skill` rows. The existing test `test_owner_can_delete_their_organization` has no skills or agents, so nothing catches it.

*Fix:* switch to `NO ACTION` (Postgres default). It gives identical protection for the direct `DELETE FROM skill_version` that `delete_version_by_id` relies on, but defers the check to end-of-statement so a sibling cascade can satisfy it. Add an integration test that deletes an org holding a custom skill pinned by a soft-deleted agent.

api/domains/agents/service.py
```diff
@@ -636,6 +704,12 @@ def create_agent(self, data: AgentCreate, context: CurrentUserContext) -> AgentR
                     detail=f"At least one of these template skills must be included in skill_ids: {', '.join(names)}",
                 )
 
+        # Preflight skill version pins before any persistence so an invalid pin
+        # (nonexistent version, pin for a skill not in the request) is rejected
+        # atomically rather than after the agent, configs, and secrets are saved.
+        resolved_pins = self._resolve_skill_pins(data.skill_ids, data.skill_versions, set(), [])
```
aai-labs-agent: ⚠️ **High — Cross-org version-existence oracle (from @Dominykas's review, verified)**

`_resolve_skill_pins` (here in `create_agent`, and at line 1587 in `update_agent`) runs **before** the accessibility check (`_resolve_skills` at line 849, `_validate_skill_update` at line 1751). `get_version(skill_id, pin.version)` is not org-scoped, so for a known skill UUID belonging to another org:

- If the version doesn't exist → `"Version 7 not found for skill X"`
- If the version exists → no error (proceeds, then the accessibility check rejects later)

Both 404, but the message distinguishes them — an information oracle that bypasses tenant scoping. The AC explicitly requires tenant scoping at the repository level.

*Fix:* scope the preflight's lookups to skills accessible to the caller's org (reuse `find_accessible_for_org`, loaded a few lines later), or move the accessibility check ahead of the pin preflight while keeping both before any persistence.

api/domains/agents/service.py
```diff
@@ -1768,11 +1863,14 @@ def update_agent(self, agent_id: UUID, data: AgentUpdate, context: CurrentUserCo
                         actor_display=secret_actor_display,
                     )
 
-        # Apply skill changes
+        # Apply skill changes (pins were preflighted before persistence above)
         for skill_id in data.removed_skill_ids:
             self.repository.remove_skill(agent.id, skill_id)
-        for skill_id in dict.fromkeys(data.skill_ids):
-            self.repository.add_skill(agent.id, skill_id)
+        for pin in resolved_skill_pins:
```
aai-labs-agent, line 1979: ⚠️ **Medium — Skill in both `skill_ids` and `removed_skill_ids` silently dropped (from @Dominykas's review, verified)**

Removals apply first (`remove_skill` at line 1867), then `re_pin_skill` (line 1871) returns silently when the row is gone (repository returns `None` if `row is None`). The previous code called `add_skill` unconditionally, re-adding the skill.

Not reachable from the current UI, but a silent API behaviour change — either restore the old behaviour (re-add) or reject the contradictory request explicitly with a 400.

api/domains/skills/service.py
```diff
@@ -38,101 +40,138 @@ def _org_id(self, context: CurrentUserContext) -> UUID:
...
-        except zipfile.BadZipFile:
-            raise HTTPException(
-                status_code=status.HTTP_400_BAD_REQUEST,
-                detail="Uploaded file is not a valid zip archive",
-            )
+            return validate_files([(f.path, f.content) for f in files], entry_path=entry_path)
+        except ValueError as exc:
+            raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail=str(exc)) from exc
+
+    def _allocate_slug(self, name: str, org_id: UUID) -> str:
```
aai-labs-agent, line 49: ⚠️ **Medium — Read-then-write race on slug allocation (from @Dominykas's review, verified)**

`_allocate_slug` and `_allocate_fork_name` (line 65) do a SELECT to find taken slugs, then use the result to pick a unique slug — without a transaction lock or retry. Two concurrent creates in one org pick the same slug; the loser hits `uq_skill_organization_slug` as an unhandled `IntegrityError` → 500.

*Fix:* wrap `create_skill`/`fork_skill` in a retry on `IntegrityError` (re-allocate and retry once), or catch and re-raise as 409.

api/domains/skills/service.py
```diff
@@ -142,65 +181,236 @@ def update_skill(self, skill_id: UUID, data: SkillUpdate, context: CurrentUserCo
...
                 status_code=status.HTTP_403_FORBIDDEN,
-                detail="Cannot delete built-in skills",
+                detail="Cannot modify built-in skills",
             )
-        if self.repository.is_assigned_to_any_agent(skill_id):
+        return skill
+
+    def get_skill_draft(self, skill_id: UUID, context: CurrentUserContext) -> SkillDraftRead:
+        org_id = self._org_id(context)
+        skill = self._get_or_404(skill_id, org_id)
```
aai-labs-agent: ⚠️ **Medium — `GET /{skill_id}/draft` doesn't reject built-ins (from @Dominykas's review, verified)**

`get_skill_draft` uses `_get_or_404` + `SKILL_MANAGE` instead of `_require_draftable_skill`, so a built-in skill returns 404 ("No draft for skill X") where every other draft operation returns 403 ("Cannot modify built-in skills"). The AC says all draft operations are "rejected for built-ins" — contractually inconsistent, though behaviorally harmless.

*Fix:* replace `_get_or_404` + `require_organization` with `_require_draftable_skill` for consistency.

ui/src/features/skills/components/skill-detail-page.tsx
```diff
@@ -0,0 +1,603 @@
...
+    setDescription(detail.description ?? "");
+    setSelectedProviders(detail.requiredProviders);
+  }
+  if (editing && existingDraft && localFiles !== null && !draftApplied) {
+    const draftFiles = existingDraft.files.map((f) => ({ path: f.path, content: f.content }));
+    setLocalFiles(draftFiles);
+    setDraftApplied(true);
+  }
+
+  const latestVersion = Math.max(0, ...versions.map((v) => v.version));
```
aai-labs-agent: 💡 **Low — "v0" flash (from @Dominykas's review, verified)**

`Math.max(0, ...versions.map(...))` renders `v0` in the page header and "Published v0" toggle while `useSkillVersions` is pending. `detail.version` is already loaded and correct — use it as the fallback instead of computing from the not-yet-loaded versions array.

Same pattern in `skill-version-history.tsx:22`.

ui/src/features/agents/components/agent-skills-tab.tsx
```diff
@@ -69,7 +79,7 @@ export const AgentSkillsTab = forwardRef<
   ];
 
   const hasPendingChanges =
-    pendingAddIds.length > 0 || pendingRemoveIds.length > 0;
+    pendingAddIds.length > 0 || pendingRemoveIds.length > 0 || Object.keys(pendingPins).length > 0;
```
aai-labs-agent: 💡 **Low — Spurious dirty state (from @Dominykas's review, verified)**

`hasPendingChanges` counts any entry in `pendingPins` (`Object.keys(pendingPins).length > 0`), so re-selecting the version the agent is already pinned to enables Apply. Compare against `skill.version` before marking dirty — only include pins where the selected version differs from the current one.

kalibetre:
## Review fixes posted

Updated the PR branch with the requested fixes in `61a73052`, followed by `be2d62bf` to merge the current `staging` base and keep the migration graph self-contained.

- Fixed the Alembic merge parent and verified a single migration head.
- Changed pinned-version foreign keys to `NO ACTION` so direct version deletion remains protected while organization cascades work.
- Added tenant-scoped skill-pin validation, overlap rejection, and cross-organization regression coverage.
- Batched pinned skill-version file loading and removed the draft null assertion.
- Restricted `PATCH /skills/{id}` to names; draft metadata now distinguishes omitted values from explicit `null`.
- Added built-in draft access protection and duplicate-name conflict handling.
- Fixed UI skill renaming, version `v0` fallback, and pin dirty-state tracking.
- Added API and Playwright regression coverage and removed unused repository methods.

Verification completed:
- `make check-migrations`
- `make check-api`
- `make lint-ui`
- `make check-ui`
- 323 focused API integration tests passed
- 12 skill-detail Playwright tests passed

ui/src/features/skills/components/skill-detail-page.tsx
```diff
@@ -0,0 +1,624 @@
...
+      prev.includes(value) ? prev.filter((p) => p !== value) : [...prev, value],
+    );
+  }
+
+  async function enterEditing() {
+    try {
+      const draft = await startDraft.mutateAsync(skillId);
+      setLocalFiles(draft.files);
+      setName(detail!.name);
+      setDescription(detail!.description ?? "");
```
aai-labs-agent, line 182: ⚠️ **Draft metadata not hydrated from the draft on editor re-entry**

`enterEditing()` seeds description and required-providers from the skill lineage (`detail`), but the `startDraft` response (a `SkillDraft`) carries the draft's own staged `description` and `requiredProviders`. If a user saves a draft with different metadata, exits, and re-enters, the editor shows the lineage's old values. Saving again silently overwrites the draft's staged metadata — data loss with a success toast.

The draft's metadata is already surfaced in the read-only preview (lines 556, 582 — `existingDraft?.description`, `existingDraft?.requiredProviders`), so the data is available — the editor seeding just doesn't use it.

Suggested fix:
```ts
setDescription(draft.description ?? "");
setSelectedProviders(draft.requiredProviders);
```

The same applies to the auto-open hydration at line 96–97, which seeds from `detail` instead of `existingDraft`.

api/domains/skills/repository.py
```diff
@@ -21,6 +31,282 @@ class SkillRepository:
...
+            return {}
+        with Session(self.delegate.engine) as session:
+            query = (
+                select(SkillVersion.skill_id, func.max(col(SkillVersion.version)))
+                .where(col(SkillVersion.skill_id).in_(skill_ids))
+                .group_by(col(SkillVersion.skill_id))
+            )
+            return {skill_id: version for skill_id, version in session.exec(query).all()}
+
+    def get_version(self, skill_id: UUID, version: int) -> SkillVersion | None:
```
aai-labs-agent, line 75: 💡 `get_version` is not org-scoped — it queries by `skill_id` and `version` without filtering by `organization_id`.

In `_resolve_skill_pins`, the accessibility check (`find_accessible_for_org`) runs first, so only accessible skills reach `get_version`. This is defense-in-depth acceptable, but adding an `org_id` parameter (matching the pattern in `get_by_id_for_org` and `is_skill_version_pinned`) would make the intent explicit and guard against future call sites that skip the pre-check.

aai-labs-agent (commented):
## Hermes Agent Review — AF-261 Skills v2

**Verdict: Comment** (1 warning, 2 suggestions — nothing blocking)

### Scope

63 files, +6236/-1163. Replaces the upload-and-replace ZIP model with draft/publish authoring, built-in forking, per-version deletion, and agent version pinning. Reviewed against the Jira AC (AF-261) and the repo's AGENTS.md review protocol.

### Migration head integrity ✅

Verified the full staging migration graph (75 revisions). Staging has 1 head (`d5e2a9f4c781`). After adding the PR's 6 migrations, there is exactly 1 head (`04088157c78c`). The chain is valid — no multiple heads at deploy time.

### AC cross-check

| AC | Status | Evidence |
|---|---|---|
| Draft/publish authoring (one draft per skill) | ✅ | `uq_skill_draft_skill_id` unique constraint; `POST/PATCH/DELETE /draft` + `/draft/publish` |
| Member cannot mutate drafts; built-ins reject drafts | ✅ | `_require_draftable_skill` checks `SKILL_MANAGE` + rejects `AAI_CLI` |
| Reload after fork hydrates from persisted draft | ✅ | `?edit=1` auto-opens editor; `useSkillDraft` fetches draft; files hydrated from draft |
| Built-in fork creates org-scoped copy + opens draft | ✅ | `fork_skill` in service.py; carries over description, providers, entry path |
| Fork name disambiguation | ✅ | `_allocate_fork_name` with `(fork)` suffix |
| Per-version deletion (last + pinned versions protected) | ✅ | `delete_skill_version` checks `len(versions) <= 1` and `is_skill_version_pinned`; composite FK safety net |
| UI shows pinned versions; disables delete for pinned | ✅ | `SkillVersionRead.is_pinned_by_agent`; version history UI |
| Whole-lineage deletion removed | ✅ | `delete_skill` replaced by `delete_skill_version` |
| Agent pins exact version at assign/update | ✅ | `AgentSkill.pinned_version`; `skill_versions` in `AgentCreate`/`AgentUpdate` |
| Publishing newer version doesn't move existing pin | ✅ | `publish_version`/`publish_draft` never touch `AgentSkill` rows |
| Re-pin from agent configuration page | ✅ | `Select` component in `agent-skills-tab.tsx` |
| Omitted skills pin to latest | ✅ | `_resolve_skill_pins` defaults to `get_latest_version` |
| Invalid pin rejected before any persistence | ✅ | `_resolve_skill_pins` called before agent/config/secret saves |
| Removing skill after changing version doesn't leave stale pins | ✅ | `markForRemoval` cleans `pendingPins`; save handler filters removed skills |
| Clearing description clears it server-side | ✅ | `SkillDraftUpdate` uses `model_fields_set`; `handleSaveDraft` sends `null` for empty description |
| Tenant-scoped draft operations (repository level, 404 not 403) | ✅ | `get_by_id_for_org` conceals other orgs; returns 404 |
| Batch skill-loading excludes soft-deleted agents | ✅ | `get_skills_for_agents_with_versions` joins `Agent.deleted_at IS NULL` |
| DB-level constraint prevents deleting pinned version | ✅ | Composite FK `fk_agent_skill_pinned_version` with `NO ACTION` |
| Agent start logs warning for missing pinned files | ✅ | Warning in `start_agent` when `pin.skill_id not in files_by_skill` |
| Legacy backfill validates paths + size/count limits | ✅ | `_readable_entries` + `_normalize_path` in migration; inlined from `files.py` |
| Feature doc + ADRs updated | ✅ | `templates-and-skills.md` + `agents.md` + 2 ADRs |
| Epic CHANGELOG | N/A | `templates-and-skills.md` is a single file, not a directory |

### ⚠️ Warning

**Draft metadata not hydrated on editor re-entry** (`ui/src/features/skills/components/skill-detail-page.tsx:182`)

`enterEditing()` seeds the editor's description and required-providers from the skill lineage (`detail`), not from the returned draft. If a user saves a draft with different metadata, exits, and re-enters the editor, the editor shows the lineage's old metadata. Saving again silently overwrites the draft's staged metadata with the lineage's values — data loss with a success toast.

The draft's `description` and `requiredProviders` are available in the `SkillDraft` response (and are already used in the read-only draft preview at line 556/582), but the editor seeding ignores them.

### 💡 Suggestions

1. **`get_version` is not org-scoped** (`api/domains/skills/repository.py:75`) — Called during `_resolve_skill_pins` after the accessibility check (`find_accessible_for_org`), so defense-in-depth is present. Adding an `org_id` parameter would make the intent explicit.

2. **`add_skill` / `re_pin_skill` commit individually** (`api/domains/agents/repository.py`) — In the `update_agent` loop, each call opens its own Session and commits. If the second `add_skill` fails (e.g., composite FK violation after a concurrent version delete), the first is already committed. Mitigated by the preflight in `_resolve_skill_pins`, but wrapping the loop in a single transaction would eliminate the window.

### Suppressed

- **`ondelete="NO ACTION"` on the composite FK** — checked per the skill's PostgreSQL `RESTRICT` vs `NO ACTION` guidance. `NO ACTION` is correct here: it defers the check to end-of-statement so sibling cascades (org → skill → skill_version) can satisfy it, while still blocking direct `DELETE FROM skill_version` for pinned versions.
- **`transformKeys` casing** — frontend Zod schemas use camelCase (`isAssignedToAgent`, `rootDir`, `hasDraft`), backend uses snake_case. Confirmed `transformKeys: true` is the default in the API client factory — this is correct, not a mismatch.
- **Helm chart version** — no `helm/` files changed; no bump needed per `operations.md`.
- **`SkillUpdate` is name-only with `extra="forbid"`** — intentional: content metadata changes go through draft/publish, not PATCH.

### Looks good

- Pin preflight before any persistence (atomic rejection)
- `SkillDraftUpdate` uses `model_fields_set` to distinguish omitted from null
- `update_skill` uses `exclude_unset=True`
- Backfill migration handles bad zips, binary content, and empty archives gracefully
- Seeder publishes a new version only when content differs (no version inflation on restart)
- Manifest builder handles root_dir collisions with first-claim-wins + logging
- TanStack Query cache invalidation uses `skillsKey.all` prefix — covers detail, draft, versions
- Version selector in agent-skills-tab properly cleans up `pendingPins` on removal

---
*Reviewed by Hermes Agent*

PR #129, opened 2026-08-17
AF-262: Google Workspace integration for agents via gogcli

sami-bre:
Gives agents Gmail, Calendar, Drive and Sheets access through the [gog CLI](https://gogcli.sh) (gogcli v0.37.0) under a single OAuth consent, instead of one consent per service. Design doc: `docs/plans/gogcli-google-workspace-integration.md`.

## Approach

**Consent stays in the web UI.** Agent pods are headless and must never see the user's Google credentials, so the existing popup flow is reused — but scopes are now derived per request from the selected services plus a read-only toggle, mirroring gog's own service map at the pinned version. The flow also requests the identity scopes so `/token` can report the account email (gog keys stored tokens by email) and returns the scopes Google *actually* granted, since the consent screen lets a user withhold individual ones.

**The platform keeps custody of the refresh token.** At start the pod receives the credential as `GOG_*` environment plus a secret-free `gog-setup.sh` that installs the OAuth client and imports the token into a file keyring. That state is wiped and rebuilt every boot, so the encrypted row stays the single source of truth, the keyring password rotates per start, and removing the credential removes pod access at the next restart with nothing left behind.

**A user-owned Web OAuth client is the expected setup, not an option.** Google caps apps requesting unverified sensitive scopes at 100 grant-users for the lifetime of the project, and that cap cannot be reset — so a server-owned client only suits a verified or Workspace-internal deployment. The UI walks through creating one and reads the client id/secret out of the downloaded JSON.

Agent guidance ships only as an AGENTS.md block; no skill file is seeded.

## Testing

- `make check-api`, `make check-ui`, `make check-migrations` clean; **1440 API tests pass** (`make test-api`).
- New: `tests/unit/test_gog_artifacts.py` (19 cases), workspace scope-derivation and email/granted-scopes cases in `test_google_oauth.py`, validator cases incl. scope revocation and identity drift, content-model cases, and start-agent materialization tests covering both runtimes' `GOG_HOME`, the absent-credential case, and the aai-cli policy gating.
- Verified end-to-end in containers: the exact Dockerfile layer builds, and the generated `gog-setup.sh` installs the client and imports the token as the unprivileged user on both `node` and `hermes`.

## Review notes

- **Base images must be rebuilt and published before this is useful** — `gog` is baked in, and both `VERSION` files are bumped (openclaw `0.4.5`, hermes `0.1.5`).
- The scope map in `google_oauth/routes.py` is pinned to gog v0.37.0's derivation; it needs review whenever `GOG_VERSION` moves.
- Roughest UX edge, accepted for v1: `services` and `readOnly` can be edited after connecting without forcing a reconnect. The field hint warns, and the validator surfaces the resulting scope drift.


api/migrations/versions/0e48aaf59368_merge_google_workspace_and_staging_heads.py
```diff
@@ -0,0 +1,23 @@
...
+
+"""
+
+from collections.abc import Sequence
+
+# revision identifiers, used by Alembic.
+revision: str = "0e48aaf59368"
+down_revision: str | Sequence[str] | None = ("b7e2c4d81f39", "d5e2a9f4c781")
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
```
aai-labs-agent, line 15: 🔴 **B1 — Deploy blocker: stale migration head**

`down_revision = ("b7e2c4d81f39", "d5e2a9f4c781")` — `d5e2a9f4c781` is no longer a staging head. Staging advanced since the branch point (`eb664e8e` → `df88b919`), and migration `04088157c78c` was added, consuming `d5e2a9f4c781`.

After this PR's migrations are applied, there are **2 heads** (`04088157c78c` + `c9f1b30a7d42`). Alembic will fail at deploy with *"Multiple heads are present; please specify which head to begin from."*

**Fix:** Rebase onto current staging. Update this merge's `down_revision` to reference `04088157c78c` (the current staging head) instead of `d5e2a9f4c781`.

ui/src/features/agents/components/integration-fields.tsx
```diff
@@ -48,91 +164,24 @@ export function IntegrationFields({
...
       )}
 
+      {renderedFields}
+
       {provider.authMethod === "google_oauth" && onOAuthConnected && (
         <GoogleAuthButton
           connected={isOAuthConnected(draft)}
           onConnected={onOAuthConnected}
-          disabled={disabled}
+          disabled={disabled || (needsServices && services.length === 0)}
```
aai-labs-agent: 🔴 **B2 — services/read_only can diverge from consented scopes (blocking)**

`disabled` is tied to agent-run state (see line 20 comment), not OAuth-connected state. After `onOAuthConnected` returns, `services` (checkbox-list) and `readOnly` (radio) stay editable. A user can: tick Gmail → Connect (button enables on services alone, `readOnly` unset → `read_only=false` → full Gmail scopes) → then tick Sheets → select Read-only → Save. Stored content diverges from granted scopes.

`GOG_TOKEN_JSON` then declares `sheets` to gog with no Sheets/Drive scope in the token, and `build_gog_policy_md` tells the agent it has Sheets *and* that the credential is read-only. The validator can't catch this — it compares `content.scopes` against a refresh grant that always returns the same scopes (tautology).

**Fix (API side):** Add a `model_validator(mode="after")` on `GoogleWorkspaceContent`: when `scopes` is non-empty, every declared service's scope set at the declared `read_only` level must be a subset of `scopes`.

**Fix (UI side):** On `onOAuthConnected`, lock the services/read-only controls until the user hits "Reconnect Google account"; or clear `refreshToken` on change so the button falls back to "Authenticate with Google".

---

⚠️ **M1 — `read_only` doesn't gate the Connect button**

`disabled` only checks `services.length === 0`. Drafts start as `content: {}`, so `readOnly` is undefined until clicked. Consent happens with an implicit `read_only=false`. Add `readOnly` to the disabled condition, or seed the draft with `readOnly: "false"`.

ui/src/features/agents/components/hire-dialog-primitives.tsx
```diff
@@ -259,6 +343,35 @@ export function GoogleAuthButton({
...
+        <GoogleGlyph size={15} />
+        {isConnecting
+          ? "Waiting for Google…"
+          : connected
+            ? "Reconnect Google account"
+            : "Authenticate with Google"}
+      </button>
+      {connected && !isConnecting && (
+        <span className="text-[0.75rem] font-medium" style={{ color: "var(--ok, #2f855a)" }}>
+          ✓ Connected — {connectedNote}
```
aai-labs-agent: ⚠️ **G1 — AC2: connected account email is never displayed**

The email is captured (`GoogleOAuthResult.email`) and stored with the credential, but the connected line shows a static `✓ Connected — {connectedNote}` where `connectedNote = "Google account connected"`. The AC says "UI shows... connected account email."

**Fix:** Show `draft.content.email` in the connected line — e.g. `✓ Connected as user@example.com`. Cheapest path: pass the email through to `connectedNote`, or render it from the draft next to the button.

api/domains/agents/models.py
```diff
@@ -168,28 +170,39 @@ class BitbucketContent(_RepoListCompat):
...
 
+    ``client_id``/``client_secret`` are optional: empty means the
+    server-owned client is backfilled from config at agent-start time.
+    """
 
-class GoogleSheetsContent(SecretContent):
-    # Same shape and rationale as GmailContent: the "Authenticate with Google" flow
-    # stores only the refresh token and the app-owned client id/secret are injected
-    # from config at agent-start time. A user's own Google client carries all three.
+    email: str
```
aai-labs-agent: ⚠️ **M2 — `email` accepts `""`**

`email: str` has no length constraint; only the UI's `requireEmail` guard stops a blank. A blank email lands in `GOG_TOKEN_JSON["email"]`, which is the key gog stores the token under.

**Fix:** `email: str = Field(min_length=1)`.

api/domains/agents/gog_artifacts.py
```diff
@@ -0,0 +1,167 @@
...
+# an "installed" or a "web" wrapper key (internal/config/credentials.go). The platform
+# consent flow uses a Web-application client, so that is what we synthesize.
+_CLIENT_JSON_KEY = "web"
+
+# gog buckets tokens per named OAuth client; the unnamed default bucket is what a plain
+# `gog ...` invocation reads (internal/config.DefaultClientName).
+_DEFAULT_CLIENT_NAME = "default"
+
+# Human-facing labels and one worked example per service, for the agent policy block.
+_SERVICE_GUIDE: dict[str, tuple[str, str]] = {
```
aai-labs-agent, line 40: ⚠️ **M3 — Three parallel service maps with no drift guard**

`_SERVICE_GUIDE` here, `GOOGLE_WORKSPACE_SERVICES` in `models.py:101`, and `_WORKSPACE_SERVICE_SCOPES` in `routes.py` are keyed by the same service names with nothing asserting they agree. Missing one on a future addition silently drops the service from the AGENTS.md block (line 134 filters unknown keys) or 400s at authorize.

**Fix:** Add one test asserting the three key sets are equal.

---

💡 **L2 — Dead docstring path**

Line 9 points at `docs/plans/gogcli-google-workspace-integration.md`, which does not exist in the repo. Drop or add it.

api/domains/agents/service.py
```diff
@@ -2251,9 +2252,12 @@ def start_agent(self, agent_id: UUID, context: CurrentUserContext) -> AgentRead:
             assert ciphertext is not None
             decrypted[provider] = decrypt_content(provider, ciphertext, key)
         self._backfill_google_client_credentials(decrypted)
-        for google_provider in (SecretProvider.GMAIL, SecretProvider.GOOGLE_SHEETS):
+        # Only google_workspace is materialized; the retired per-service Google providers
+        # are tombstones (see SecretProvider) and a leftover secret is simply ignored here
```
aai-labs-agent: 💡 **L1 — "Tombstones" comment contradicts the code**

Migration `c9f1b30a7d42` deletes the retired providers outright — enum members, content models, and rows. `SecretProvider("gmail")` now raises `ValueError` (tested in `test_retired_google_providers_are_gone`). But this comment says they "are tombstones," and two other places say the same: `docs/features/integrations.md:13` and `ui/src/features/agents/integrations.ts:117`.

Rewrite all three to say the providers and their rows are gone and affected agents must reconnect.

---

💡 **L4 — Single-element tuple loop**

`for google_provider in (SecretProvider.GOOGLE_WORKSPACE,):` is a leftover from the multi-provider version. Collapse to a straight `isinstance` check. Same pattern at line 2801 in `_backfill_google_client_credentials`.

api/domains/integrations/google_oauth/routes.py
```diff
@@ -41,25 +41,92 @@
...
-DEFAULT_PROVIDER = SecretProvider.GMAIL.value
+
+# gog appends these to every authorization; we need them too, because the account email
+# is read out of the resulting id_token (gog keys its stored tokens by email).
+_IDENTITY_SCOPES: tuple[str, ...] = ("openid", "email", f"{_SCOPE_PREFIX}userinfo.email")
+
+# Providers a signed state may name.
+_VALID_PROVIDERS: frozenset[str] = frozenset({SecretProvider.GOOGLE_WORKSPACE.value})
+
+GOOGLE_USERINFO_ENDPOINT = "https://openidconnect.googleapis.com/v1/userinfo"
```
aai-labs-agent: 💡 **L3 — Dead constant `GOOGLE_USERINFO_ENDPOINT`**

Defined here but never referenced anywhere in the file. The `_email_from_id_token` function extracts email from the id_token via `jwt.decode`, and the validator (`google_workspace.py`) has its own private `_USERINFO_ENDPOINT`. Remove it.

aai-labs-agent (changes requested):
## Code Review — AF-262: Google Workspace integration via gogcli

**Combined review (Hermes Agent + Dominykas).** 47 files, +1868/−1453, base `staging`.

**Verdict: Changes Requested** — 2 blockers, 2 AC gaps, 5 medium/low findings.

---

### 🔴 Blockers

**B1 — Migration head chain is broken (deploy blocker)**
The merge migration `0e48aaf59368` references `d5e2a9f4c781` as a parent, but staging has advanced since the branch point (`eb664e8e` → `df88b919`). Migration `04088157c78c` was added to staging, consuming `d5e2a9f4c781` — it is no longer a head. After the PR's migrations are applied, there are **2 heads** (`04088157c78c` + `c9f1b30a7d42`). Alembic will fail at deploy with *"Multiple heads are present."* The `mergeable_state: dirty` corroborates this.

**Fix:** Rebase onto current staging. Update the merge's `down_revision` to reference `04088157c78c` (the current staging head) instead of `d5e2a9f4c781`. The `_OLD`/`_NEW` check constraint snapshots are correct — no constraint-modifying migration was added between the branch point and current staging.

**B2 — Stored `services`/`read_only` can diverge from consented scopes**
`services` (checkbox-list) and `readOnly` (radio) stay editable after the OAuth popup returns — `integration-fields.tsx` passes `disabled` tied to agent-run state, not OAuth-connected state. A user can: tick Gmail → Connect (button enables on services alone, `readOnly` unset → `read_only=false` → full Gmail scopes granted) → then tick Sheets → select "Read-only" → Save. Stored content is `services=[gmail,sheets]`, `read_only=true`, `scopes=[gmail.modify, …]`.

Consequences: `GOG_TOKEN_JSON` declares `sheets` to gog with no Sheets/Drive scope in the token, and `build_gog_policy_md` tells the agent it has Sheets *and* that the credential is read-only. The agent advertises capabilities it does not have and mis-describes its own access level. The reverse (consent read-only, store full) makes every write fail at Google.

`validate_google_workspace` cannot catch this: it compares `content.scopes` against the freshly granted scope string from a refresh grant — and `content.scopes` was itself copied from that grant. The check is a tautology.

**Fix (API is the load-bearing half):**
- Move `_WORKSPACE_SERVICE_SCOPES` and `_IDENTITY_SCOPES` out of `routes.py` into a module `models.py` can import without a cycle (e.g. `api/domains/integrations/google_oauth/scopes.py`).
- Add a `model_validator(mode="after")` on `GoogleWorkspaceContent`: when `scopes` is non-empty, every declared service's scope set *at the declared `read_only` level* must be a subset of `scopes`; otherwise 422 naming the offending service.
- UI: on `onOAuthConnected`, lock the services/read-only controls until the user hits "Reconnect Google account"; or clear `refreshToken` on change so the button falls back to "Authenticate with Google".

---

### ⚠️ AC2 Gaps (partly unmet)

**G1 — Connected account email is never displayed**
The email is captured (`GoogleOAuthResult.email`) and stored with the credential, but `GoogleAuthButton` shows a static `✓ Connected — Google account connected`. The AC says "UI shows... connected account email."

**Fix:** Show `draft.content.email` in the connected line — e.g. `✓ Connected as user@example.com`.

**G2 — "Validate" is not reachable from the UI**
`useValidateIntegration` has zero consumers, and `AgentSecretReadSchema` carries no validation state. The backend endpoint and `validate_google_workspace` are correct and well tested — the gap is purely the missing affordance. Pre-existing (the hook predates this branch), but the ticket puts it in scope.

**Fix:** Add a Validate button per configured integration, wire it to the existing hook, and render `validationIdentity`, `validationError`, and `missingScopes`.

---

### Medium

**M1 — `read_only` does not gate the Connect button**
Drafts start as `content: {}`, so `readOnly` is undefined until clicked. `hasIncompleteIntegration` requires it (Hire stays disabled), but `GoogleAuthButton`'s `disabled` only checks `services.length === 0`. Consent happens with an implicit `read_only=false`. Add `readOnly` to the disabled condition, or seed the draft with `readOnly: "false"`.

**M2 — `email` accepts `""`**
`GoogleWorkspaceContent.email: str` has no length constraint; only the UI's `requireEmail` guard stops a blank. A blank lands in `GOG_TOKEN_JSON["email"]`, which is the key gog stores the token under. Add `Field(min_length=1)`.

**M3 — Three parallel service maps with no drift guard**
`GOOGLE_WORKSPACE_SERVICES` (models.py), `_WORKSPACE_SERVICE_SCOPES` (routes.py), and `_SERVICE_GUIDE` (gog_artifacts.py) are keyed by the same names with nothing asserting they agree. Missing one on a future addition silently drops the service from the AGENTS.md block or 400s at authorize. Add one test asserting the three key sets are equal.

---

### Low

**L1 — "Tombstones" documentation contradicts the code**
Migration `c9f1b30a7d42` deletes the retired providers outright (enum members, content models, and rows — `SecretProvider("gmail")` now raises `ValueError`, tested). But three places still describe them as surviving tombstones:
- `docs/features/integrations.md:13` — "their enum members and content models survive as tombstones"
- `ui/src/features/agents/integrations.ts:117` — "The backend keeps their enum members so existing secrets still decrypt"
- `api/domains/agents/service.py:2256` — "are tombstones (see SecretProvider)"

Rewrite all three to say the providers and their rows are gone and affected agents must reconnect.

**L2 — Dead docstring path**
`gog_artifacts.py:9` points at `docs/plans/gogcli-google-workspace-integration.md`, which does not exist in the repo. Drop or add it.

**L3 — Dead constant `GOOGLE_USERINFO_ENDPOINT`**
`routes.py:94` defines `GOOGLE_USERINFO_ENDPOINT` but it's never referenced — the validator has its own `_USERINFO_ENDPOINT`. Remove it.

**L4 — Single-element tuple loops**
`service.py:2258` and `service.py:2801` both iterate `for google_provider in (SecretProvider.GOOGLE_WORKSPACE,)` — a leftover from the multi-provider version. Collapse to a straight `isinstance` check.

**L5 — `onReposChange` naming lies for checkbox-list**
`integration-fields.tsx:86` routes `checkbox-list` changes through `onReposChange`. It works (the handler is key-generic) but the prop name now lies. Rename to `onListChange` at its call sites.

---

### ✅ Looks Good

- **Scope derivation** — `workspace_scopes()` mirrors gog v0.37.0's service map. Dedupes overlapping scopes, applies read-only variants, always includes identity scopes. Pinned-version caveat documented.
- **Token custody** — Refresh token in pod Secret (`GOG_TOKEN_JSON`), not ConfigMap. `gog-setup.sh` entirely env-driven. `umask 077`. State wiped per boot.
- **Runtime parity** — Hermes and OpenClaw builders, start scripts, and Dockerfiles are symmetric. `GOG_HOME` deliberately differs to avoid the Hermes PVC. Tested for both runtimes.
- **Enum cascade completeness** — All functions keyed on `SecretProvider` updated: `PROFILE_SLUGS`, `PROVIDER_DISPLAY_NAMES`, `PROVIDER_CONTENT_MODELS`, validators, UI labels, skill seeder. `SecretProvider("gmail")` raises.
- **UI mutation wiring** — OAuth callback writes `email` + `scopes` alongside `refreshToken` via `setFields` (batch update, fixes the previous successive-setState overwrite bug). `coerceBooleanFields` includes `readOnly`.
- **Skill seeder pruning** — `_prune_retired_aai_cli_skills` safely deletes unreferenced retired skills, leaves assigned ones with a warning.
- **Documentation sync** — `docs/features/integrations.md` comprehensively updated (beyond the tombstone wording issue). Template seeds updated.
- **Test coverage** — 19 new `test_gog_artifacts.py` cases, expanded scope-derivation/validator/identity tests, integration tests covering both runtimes, absent-credential case, and aai-cli policy gating.
- **Base images** — `VERSION` bumped in both, `smoke-test.sh` asserts `gog --version`, `gog` pinned at v0.37.0.

- **`_email_from_id_token` doesn't verify JWT signature** — intentional; token arrives directly from Google's token endpoint in a server-to-server TLS exchange. Docstring explains why.
- **`gog-setup.sh` failure is non-fatal** (`|| echo "setup failed; continuing"`) — matches the established `aai-cli-setup.sh` pattern in both start.sh scripts.
- **VERSION bumps without Helm chart changes** — per `docs/guidelines/operations.md`, base image VERSIONs are for base images. No `helm/` files changed.

---

*Combined review by Hermes Agent and Dominykas. Reviewed against Jira AF-262 acceptance criteria and the repo's AGENTS.md review protocol.*

PR #137, opened 2026-08-24
AF-271 Refactor agent communications into a pluggable platform-plugin architecture

kalibetre:
This PR restructures how agents connect to messaging platforms. Platform-specific configuration (Slack/Telegram/Discord) is extracted from the Agent model into a new **Communications** domain with first-class `CommunicationConnection` records, a separately-served gateway, and a plugin registry — so adding a future platform no longer touches the agent data model. The agent creation flow drops platform credentials entirely; connections are now managed in a revamped Messaging section of the agent UI. The local development workflow is replaced by a one-command `run.sh`/`stop.sh` flow backed by k3d, and CI's Kubernetes tests migrate from a bastion SSH tunnel to k3d-action.

## Communications domain (new)

A new `communications` domain owns everything about agent-to-platform messaging. It introduces:

- **CommunicationConnection** — an agent-owned record targeting one platform, replacing the per-platform config tables (`agent_slack_config`, `agent_telegram_config`, `agent_discord_config`). An agent can now hold multiple connections, including several on the same platform.
- **Platform Plugins** — a registry with shipped implementations for Slack, Discord, and Telegram. Each plugin owns its schema, webhook handling, and token validation. The plugin base defines the contract (settings schema, credentials schema, ingress/egress interfaces) that new platforms implement.
- **Communications gateway** — a separately-served FastAPI app (`communications_main.py`, port 8002) with three API surfaces: runtime-facing (delivery claim/complete/reply), driver-facing (connection lifecycle), and provider-facing (webhooks). An outbound processor loop and an ingress supervisor run in-process.
- **Delivery pipeline** — outbound messages flow through a `CommunicationDelivery` record that the runtime adapter claims, drives through the local runtime's chat completions API, and completes back.

The legacy `AgentPlatform` enum, per-platform config models, and their repository methods are removed. The `Agent` model drops the `platform` column (migration `8f5c1a6e4b93`). Agent creation and update schemas no longer carry platform credentials — `AgentCreate` requires only `name`, `agent_type`, and template reference.

## Runtime adapter pattern

Both OpenClaw and Hermes runtimes now communicate with the platform through a shared `communications-runtime-adapter.py` sidecar, which bridges the Communications gateway's delivery protocol to the runtime's local HTTP chat-completions API. This replaces the previous approach where each runtime embedded platform-specific logic directly.

- Builders drop all per-platform config builders (`build_openclaw_config_overlay_slack`, `build_hermes_config_discord`, etc.) in favor of a single `build_openclaw_gateway_config` / `build_hermes_gateway_config` that takes only model and LiteLLM URL.
- Runtime secrets are standardized around `RUNTIME_API_KEY`, `RUNTIME_API_URL`, and `RUNTIME_MODEL` — the same shape for both runtimes — replacing the previous mix of `SLACK_BOT_TOKEN`, `TELEGRAM_BOT_TOKEN`, `DISCORD_BOT_TOKEN`, etc.
- The Hermes healthz server drops platform token validation, gateway-state monitoring, and the `agent_slack_tokens_ok` metric; liveness is now a simple sidecar check since provider sessions run in Communications.

## Agent UI: Messaging revamp

The agent Messaging section is overhauled around the new connection model:

- Connection cards show platform icon, external identity, and a live status dot.
- The "New connection" form uses a responsive platform tile grid (3–4 columns, text-below-icon when space is tight, icon-only on very small screens) and derives the connection name from the selected platform.
- Schema fields render human-readable labels, hints, and required marks; credentials use a replace-only-with-blanks editor.
- Retiring a connection is behind a confirmation dialog.
- The agent detail banner links straight to Messaging; the tab is reachable via `?connect` to auto-open the new-connection form.
- Copy is platform-agnostic ("a messaging platform") so adding future platforms doesn't stale the UI text.

The legacy multi-step hire wizard (platform-choice, Slack/Telegram/Discord credential steps) is removed from the hire dialog; connection setup now lives in the Messaging section. Full-color brand icons replace the single-color glyphs.

## Teams plugin removal

The unsupported Microsoft Teams plugin is removed entirely — `AgentTeamsConfig`, `build_secret_teams`, `build_openclaw_config_overlay_teams`, and related UI panels. Hermes' platform-restriction check is dropped since it only guarded against Teams.

## Breaking Changes

- **`Agent.platform` removed.** Any code reading `agent.platform` or filtering agents by platform must use `CommunicationConnection` instead. The `AgentCreate` and `AgentUpdate` schemas reject platform-specific fields (`slack_bot_token`, `telegram_bot_token`, `discord_bot_token`, `teams_app_id`, etc.).
- **Per-platform config tables dropped.** `agent_slack_config`, `agent_teams_config`, `agent_telegram_config`, `agent_discord_config` are removed by migration. Existing connections are migrated to `communication_connections` by `6d3a9e4c2f71`.
- **Agent creation no longer configures a platform.** Creating an agent no longer accepts or requires Slack/Telegram/Discord credentials; connections are added separately through the Messaging section.
- **Hermes healthz metrics changed.** The `agent_slack_tokens_ok` metric and token-validation polling are removed; alerting that referenced it needs updating.
- **Runtime secret names standardized.** OpenClaw pods now expect `OPENCLAW_GATEWAY_TOKEN`/`RUNTIME_API_KEY` instead of `SLACK_BOT_TOKEN`/etc.; Hermes pods expect `RUNTIME_API_KEY` instead of platform-specific tokens.
- **`make up/down/restart/logs/build/clean` targets removed.** Use `./run.sh` and `./stop.sh` instead.


api/domains/communications/plugins/slack.py
```diff
@@ -0,0 +1,376 @@
...
+        # that timestamp as the canonical message identity so lifecycle
+        # feedback can reliably target the inbound message; the optional
+        # client-generated id remains useful metadata but is not a Slack API
+        # message reference.
+        message_id = str(event.get("ts") or "")
+        if not channel_id or not message_id:
+            return []
+        try:
+            occurred_at = datetime.fromtimestamp(float(event.get("ts", "0")), tz=UTC)
+        except TypeError, ValueError, OSError:
```
aai-labs-agent, line 294: 🔴 **Critical — Syntax error (deploy blocker)**

`except TypeError, ValueError, OSError:` is Python 2 syntax. In Python 3, multiple exception types must be parenthesized:

```python
except (TypeError, ValueError, OSError):
```

Confirmed via `ast.parse()` — the module fails to import with `SyntaxError: multiple exception types must be parenthesized`. This prevents the entire Slack plugin from loading, which would crash the communications gateway when the plugin registry tries to import it.

api/migrations/versions/6d3a9e4c2f71_add_communication_connections.py
```diff
@@ -0,0 +1,316 @@
...
+import uuid
+from collections.abc import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+from api.core.config import get_config
+from api.infrastructure.crypto import decrypt_token, encrypt_token
+
+revision: str = "6d3a9e4c2f71"
+down_revision: str | Sequence[str] | None = "04088157c78c"
```
aai-labs-agent, line 21: 🔴 **Critical — Stale `down_revision` creates a branch (deploy blocker)**

`down_revision = "04088157c78c"` — but `04088157c78c` is no longer the HEAD on `staging`. Migration `9b4c7d2e6f10_add_discord_allow_all_users.py` was added to staging with `down_revision = "04088157c78c"`, making `9b4c7d2e6f10` the current HEAD.

This PR creates a second branch from `04088157c78c`. Deploying will cause Alembic to fail: "Multiple heads are present; please specify which head to begin from."

**Fix:** Change `down_revision` to `"9b4c7d2e6f10"` (the current staging HEAD). The PR's migration 3 drops `agent_discord_config`, which must run *after* `9b4c7d2e6f10` (which adds a column to that table), so chaining from `9b4c7d2e6f10` is also the correct ordering.

api/migrations/versions/7e4b0f5d3a82_add_communication_deliveries.py
```diff
@@ -0,0 +1,108 @@
...
+        sa.Column("last_error_code", sa.String(length=100), nullable=True),
+        sa.Column("last_error_message", sa.String(length=500), nullable=True),
+        sa.Column("envelope", postgresql.JSONB(astext_type=sa.Text()), nullable=False),
+        sa.CheckConstraint("attempt_count >= 0", name="ck_communication_delivery_attempt_count"),
+        sa.ForeignKeyConstraint(["agent_id"], ["agent.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(
+            ["connection_id", "organization_id"],
+            ["communication_connection.id", "communication_connection.organization_id"],
+            name="fk_communication_delivery_connection_organization",
+            ondelete="RESTRICT",
```
aai-labs-agent, line 66: ⚠️ **Warning — `RESTRICT` vs `NO ACTION` in cascade path**

`ondelete="RESTRICT"` on the `communication_delivery → communication_connection` FK can fail during org-deletion cascades. When `DELETE organization` cascades to `agent` (CASCADE) → `communication_connection` (CASCADE) + `communication_delivery` (CASCADE), PostgreSQL evaluates `RESTRICT` *immediately* during cascade processing. If the `communication_connection` cascade fires before the `communication_delivery` cascade, the `RESTRICT` check fails because delivery rows still exist.

`NO ACTION` (the Postgres default) defers the check to end-of-statement, allowing all cascades to complete first. The codebase already uses `NO ACTION` on at least one FK (agent model), so there's precedent. Consider switching `RESTRICT` → `NO ACTION` (or omitting `ondelete` since `NO ACTION` is the default).

Same concern applies to `agent_chat_message → communication_connection` in migration `6d3a9e4c2f71` (line 280) and `communication_delivery.message_id → agent_chat_message` (line 68).

api/domains/communications/service.py
```diff
@@ -0,0 +1,201 @@
...
+        connection_id: UUID,
+        data: CommunicationConnectionUpdate,
+        context: CurrentUserContext,
+    ) -> CommunicationConnectionRead:
+        agent = self.authorization.require_action(context, agent_id, PermissionKey.AGENT_UPDATE)
+        action_scope = self.authorization.authorization_scope(context, PermissionKey.AGENT_UPDATE)
+        connection = self.repository.get_active_in_scope(connection_id, agent_id, action_scope)
+        if connection is None:
+            self._raise_not_found(connection_id)
+        assert connection is not None
```
aai-labs-agent, line 104: ⚠️ **Warning — `assert` in production code**

`assert connection is not None` is stripped under `python -O`. While the code path is safe (the None case raises via `_raise_not_found` on the previous line), this matches the established anti-pattern. Consider removing the assert or using a type-narrowing comment instead.

api/domains/communications/repository.py
```diff
@@ -0,0 +1,247 @@
...
+                .where(
+                    col(CommunicationConnection.id) == connection_id,
+                    col(CommunicationConnection.agent_id) == agent_id,
+                    col(CommunicationConnection.retired_at).is_(None),
+                    *agent_scope_predicates(authorization_scope),
+                )
+            )
+            return session.exec(query).one_or_none()
+
+    def get_active(self, connection_id: UUID) -> CommunicationConnection | None:
```
aai-labs-agent, line 72: 💡 **Suggestion — Unscoped `get_active`**

`get_active(connection_id)` queries by `connection_id` only, without an `organization_id` filter. It's only called from the gateway service (which authenticates via driver key, not user context), so there's no direct exploit. But as a defense-in-depth measure, scoping by `organization_id` would make the intent explicit and guard against future misuse from a user-facing code path.

api/domains/communications/service.py
```diff
@@ -0,0 +1,201 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="Stored Communication Connection credentials are invalid",
+            ) from exc
+
+    def _read(self, connection: CommunicationConnection) -> CommunicationConnectionRead:
+        plugin = self.plugins.require(connection.platform_key)
+        webhook_url = None
+        if PlatformCapability.WEBHOOK_INGRESS in plugin.capabilities and self.config.api_external_url:
+            webhook_url = f"{self.config.api_external_url.rstrip('/')}/communications/v1/webhooks/{connection.id}"
```
aai-labs-agent, line 193: 💡 **Suggestion — Webhook URL assumes main API ingress**

The webhook URL is built from `api_external_url` (the main API host), but the gateway routes are served on the communications app (port 8002). This works only if the modified `ingress.yaml` routes `/communications/v1/*` to the communications service. If the communications service is on a separate host, the webhook URL would be wrong. Worth verifying the ingress routing is in place.

aai-labs-agent (changes requested):
## Code Review — AF-271: Pluggable Platform-Plugin Communications Architecture

**Verdict: Changes Requested** — 2 blocking issues, 2 warnings, 2 suggestions

This is a well-architected refactoring. The new communications domain cleanly separates platform concerns from the agent model, the plugin contract is clean, and the delivery pipeline (idempotency, ordering, retry, dead-letter) is solid. The migration backfill from legacy per-platform config tables is thorough. Security — credential encryption at rest, no credentials in read APIs, driver-key/webhook authentication on the gateway, agent access scoping on the management API — is all correct.

Two blocking issues prevent merge.

---

### 🔴 Critical (Blocking)

1. **Syntax error in Slack plugin** — `api/domains/communications/plugins/slack.py:267`
   `except TypeError, ValueError, OSError:` is Python 2 syntax. In Python 3 this must be `except (TypeError, ValueError, OSError):`. Confirmed via `ast.parse()` — the module fails to import entirely. The Slack plugin cannot load, which would crash the communications gateway at startup when the plugin registry tries to import it.

2. **Stale migration `down_revision` creates a branch** — `api/migrations/versions/6d3a9e4c2f71_add_communication_connections.py:21`
   `down_revision = "04088157c78c"` — but `04088157c78c` is no longer the HEAD on `staging`. Migration `9b4c7d2e6f10_add_discord_allow_all_users.py` was added to staging with `down_revision = "04088157c78c"`, making `9b4c7d2e6f10` the new HEAD. The PR should chain from `9b4c7d2e6f10`, not `04088157c78c`. Deploying with two heads will cause Alembic to fail with "Multiple heads are present; please specify which head to begin from." This also explains the `mergeable_state: dirty`.

### ⚠️ Warnings

3. **`RESTRICT` vs `NO ACTION` on cascade-path FKs** — `api/migrations/versions/7e4b0f5d3a82_add_communication_deliveries.py:66` and `api/migrations/versions/6d3a9e4c2f71_add_communication_connections.py:280`
   `communication_delivery → communication_connection` and `agent_chat_message → communication_connection` both use `ondelete="RESTRICT"`. In an org-deletion cascade (`organization → agent (CASCADE) → communication_connection (CASCADE) + agent_chat_message (CASCADE) + communication_delivery (CASCADE)`), PostgreSQL evaluates `RESTRICT` immediately during cascade processing. If the `communication_connection` cascade fires before `agent_chat_message` or `communication_delivery`, the `RESTRICT` check fails because child rows still exist. `NO ACTION` (the Postgres default) defers the check to end-of-statement, allowing all cascades to complete. The codebase already uses `NO ACTION` on at least one FK (agent model), so there's precedent. This is reachable when org deletion proceeds after all agents are soft-deleted but their data rows remain.

4. **`assert` in production code** — `api/domains/communications/service.py:104`
   `assert connection is not None` after a None-check that raises. Under `python -O`, the assert is stripped. While the code path is safe (the None case raises via `_raise_not_found`), this matches the established anti-pattern. Use a type-narrowing `if` guard or remove the assert.

### 💡 Suggestions

5. **Unscoped `get_active` in repository** — `api/domains/communications/repository.py:72`
   `get_active(connection_id)` queries by `connection_id` only, without an `organization_id` filter. It's only called from the gateway service (which authenticates via driver key, not user context), so there's no direct exploit. But as a defense-in-depth measure, scoping by `organization_id` would make the intent explicit and guard against future misuse.

6. **Webhook URL assumes main API ingress** — `api/domains/communications/service.py:193`
   The webhook URL is constructed from `api_external_url` (the main API host). This works only if the ingress routes `/communications/v1/*` to the communications service (port 8002). Verify the modified `ingress.yaml` handles this routing — if the communications service is on a separate host, the webhook URL would be wrong.

---

### Jira AF-272 — Acceptance Criteria Cross-Check

| 1 | Agent can have 0+ connections, including multiple on same platform | ✅ | `CommunicationConnection` model allows N per agent; unique constraint is on display name, not platform |
| 2 | Slack, Telegram, Discord shipped; Teams not exposed | ✅ | Three plugins in `plugins/`; Teams config removed entirely |
| 3 | Each plugin owns settings/credential schema, validation, ingress, outbound | ✅ | `PlatformPlugin` ABC defines all contracts; `SlackPlatformPlugin`, `DiscordPlatformPlugin`, `TelegramPlatformPlugin` implement them |
| 4 | UI renders connection config from plugin schemas | ✅ | `SchemaFields` component reads `settingsSchema`/`credentialsSchema` from platform descriptors |
| 5 | Credentials encrypted at rest, never returned via read APIs, not in runtime containers | ✅ | `encrypt_token` on write; `CommunicationConnectionRead` excludes `credentials_encrypted`; runtime secrets use `RUNTIME_API_KEY` only |
| 6 | Gateway supervises connections independently from agent runtimes | ✅ | Separate FastAPI app (`communications_main.py`, port 8002) with own Helm deployment; `PlatformIngressSupervisor` manages ingress sessions |
| 7 | Deliveries are durable, idempotent, ordered, support retries + terminal failure | ✅ | `CommunicationDelivery` with idempotency key, ordering key, `attempt_count`, exponential backoff, `DEAD_LETTERED` status |
| 8 | Conversation messages retain originating connection | ✅ | `agent_chat_message.connection_id` added + backfilled in migration `6d3a9e4c2f71` |
| 9 | Connection health observed independently from agent lifecycle | ✅ | `observed_status`, `last_health_at`, `ingress_lease_*` columns; `record_health` + supervisor health loop |
| 10 | Connection access/mutations enforce Agent Access + tenant isolation | ✅ | Service uses `require_visible`/`require_action`/`require_action_for_visible` with `AGENT_UPDATE` + `AGENT_SECRET_MANAGE`; repository uses `agent_scope_predicates` |
| 11 | Legacy single-platform config + provider-specific runtime branches removed | ✅ | `agent.platform` dropped; per-platform config tables dropped; per-platform builders removed; Teams removed |
| 12 | Architecture/runtime/domain docs describe new model | ✅ | ADR added; `docs/features/communications/CHANGELOG.md` added; `docs/architecture/`, `docs/features/` updated |
| 13 | API, UI, focused e2e tests pass | ⚠️ | Slack plugin syntax error (finding #1) prevents import — tests that exercise the Slack plugin would fail. Other tests likely pass. |
| 14 | Migration policy: hard cutover, no backward compat | ✅ | Legacy tables dropped by migration; backfill only for connections; `DELETE FROM agent_chat_message WHERE connection_id IS NULL` in migration 3 |

---

- **`assert isinstance(...)` in plugin methods** (slack.py:110, 126, 135, etc.) — common pattern for type narrowing after receiving base types from the ABC. Under `-O` the asserts are stripped but the plugin registry only calls these methods with correct types. Matches established pattern.
- **Helm chart `appVersion` missing** — Chart has never had `appVersion` (4-line `Chart.yaml`); relies on `image.tag` instead. Matches repo convention. Chart `version` was bumped 0.7.8 → 0.7.9.
- **`uq_communication_connection_credential` allows NULL scope/fingerprint** — The backfill sets both fields for all connections; the unique constraint on nullable columns is standard Postgres (NULLs are distinct).
- **Webhook `Content-Encoding`** — No SSE/proxy routes in this PR; N/A.
- **UI mutation wiring** — All form fields rendered in `agent-channel-settings.tsx` are sent by the save handler; create/update/retire hooks correctly call the right endpoints with correct payloads. No silent data loss.

---

*Reviewed by Hermes Agent — 190 files, 7,516 additions, 13,421 deletions*

PR #143, opened 2026-08-26
AF-186 Skill Library

kalibetre:


dominykas-aai-labs:
## Review: AF-186 â€” scoped Skill library

Reviewed `origin/staging...HEAD` (~7.7k added lines). Diffing against `main` pulls in the unrelated communications work, so all references below are against the staging merge base.

**Overall the domain model is well-shaped.** Append-only versions with "latest == current" (so there's no current-version pointer to keep in sync), the composite FK from `agent_skill(skill_id, pinned_version)` â†’ `skill_version` giving DB-level protection for pins, the single-draft-per-lineage slot, and centralising path validation in `api/domains/skills/files.py` are all good calls. 86 integration tests on the skills surface is solid coverage.

The defects below are concentrated in the newly added **version-pinning paths for templates and overrides**, which are the least-tested part of the change.

---

### Blocking

#### 1. `500` when a required Skill has no published version

`api/domains/templates/service.py:118` and `:122` index `versions[skill_id]`, where `versions` comes from `get_latest_version_numbers()` â€” which only returns entries for skills that actually have a `skill_version` row.

This branch makes draft-only lineages the *normal* state: `create_skill`, `create_platform_skill`, `create_agent_skill` and both fork paths all create a Skill with a draft and **no v1**. `_validate_skill_ids` happily accepts such a skill, so any template create/update or platform-draft create/update that requires it raises an unhandled `KeyError` â†’ 500.

Same cause, same fix needed in two more places:
- `api/domains/agents/service.py:1064` and `:1069` (`latest_versions[skill_id]`), in the override-draft path.
- `api/domains/templates/repository.py` â†’ `_resolve_skill_versions` raises a bare `ValueError` for the same condition, which would also surface as a 500.

Suggest rejecting unpublished skills at the boundary with a 400/422 ("Skill X has no published version to require"), from one shared helper.

#### 2. `required_skill_versions` is rejected whenever standalone and grouped requirements coexist

`update_template` (`templates/service.py:277`, `:283`) and `update_draft` (`:570`, `:580`) each call `_resolve_skill_map` **twice** â€” once with only the standalone ids, once with only the group ids â€” but hand the *same* `required_skill_versions` dict to both. The guard at `:100`:

```python
unknown_version_ids = set(requested_versions) - set(all_ids)
```

then 422s every version entry belonging to the other half. So this always fails:

```json
{
  "required_skill_ids": ["A"],
  "required_skill_groups": [{"group_key": "g1", "skill_ids": ["B", "C"]}],
  "required_skill_versions": {"A": 1, "B": 2}
}
```

Only `create_template` (a single call) works. The one existing test (`test_templates.py:867`) uses a single standalone skill, so this is uncovered. Fix: resolve standalone + groups in one call, or split `requested_versions` per half before validating.

#### 3. Slug allocation ignores the other scopes, so mount directories can collide

`root_dir = slug`, and all three tiers mount into one flat `./skills/` tree â€” but:

- `_allocate_slug` (`skills/service.py:60`) checks only `find_org_scoped(org_id)`; platform slugs are invisible to it.
- `_allocate_platform_slug` (`:167`) checks only `find_all_global()`; org slugs are invisible to it.
- `_allocate_agent_slug` (`:330`) gets it right â€” it uses `find_visible_for_agent`.

The partial unique indexes are per-tier, so the DB won't catch it either. A Platform Skill and an Organization Skill can therefore both own `./skills/foo/`. When an agent pins both, `build_skills_manifest` (`aai_cli_skills/__init__.py:158-166`) detects the collision, **drops the loser's files**, and only logs a warning (`agents/service.py:1951`) â€” while `TOOLS.md` still points both pointers at the same path.

`find_accessible_for_org(org_id)` already exists and is the right input for the org case.

#### 4. Renaming a bundled Platform Skill breaks API startup

`update_platform_skill` (`skills/service.py:204`) has no `source == AAI_CLI` guard â€” unlike `update_skill`, which explicitly refuses to modify built-ins. But `seed_aai_cli_skills` dedups on **name** (`skill_seeder.py:31`) and then inserts with the hardcoded bundle slug.

So: rename "Jira" â†’ "Jira (legacy)"; on the next API boot the seeder no longer finds "Jira", inserts a new Skill with `slug="aai-jira"`, and violates `uq_skill_platform_slug`. The `IntegrityError` is uncaught and `api_app.py:68` re-raises â€” **the app fails to start.**

Dedup the seeder on `slug` (documented as the immutable identity) rather than `name`, and/or block renames of `AAI_CLI`-sourced lineages.

#### 5. Template version pins are enforced as exact equality, with no UI affordance

`_validate_required_skill_versions` (`agents/service.py:414-425`) rejects any pin that isn't `== required_version`. Meanwhile `AssignedSkillRow` in `ui/src/features/agents/components/agent-skills-tab.tsx` renders the version `<Select>` with `disabled={isRunning}` only â€” nothing marks required skills as version-locked.

So a user can pick v3 for a skill the template pins at v2 and get a 400 on save. More significantly, a template-required skill can never be upgraded on an agent without republishing the template â€” which sits awkwardly next to the AC *"Users can â€¦ select an exact published version when assigning a Skill to an Agent."*

Worth deciding which semantic you want, then making both ends agree: either relax to `>=` (template states a minimum), or disable the picker for `skill.required` rows and say why.

---

### Performance

**N+1 on the agent list endpoint.** `_build_agent_read` (`agents/service.py:505-524`) now calls, *per assigned skill*, `get_latest_version()` (1 query) and `has_update_for_pin()` (1 + up to 3 more, since `has_source_update` opens its own session and runs three selects). `list_agents` (`:1450`) calls this for every agent on the page â€” a 15-agent page with 5 skills each is ~300 round trips where it used to be zero.

`SkillRepository` already has the batch primitives (`get_latest_versions`, `get_source_update_skill_ids`) that `SkillService._to_reads` uses. Pre-load once in `list_agents` and thread the maps down.

**Migration issues 40 full-table UPDATEs per skill.** `_normalize_template_references` (migration `b6c7d8e9f0a1`, called at line 306) rewrites 8 text columns across 5 template tables, and it's called *inside* the per-skill loop of `_normalize_existing_files`. With the 12 bundled skills plus tenant skills that's hundreds of full scans holding write locks on the template tables. Collect the `(old_root, old_entry, new_root)` triples first, then do one pass per column.

---

### Migration risk â€” please dry-run against a prod snapshot

1. **`skill_version = 1` backfill can violate the new FK.** `_add_version_columns` (line 185) sets `skill_version = 1` for any requirement row whose skill has no `skill_version` row at all; lines 375-383 then add a `RESTRICT` FK to `skill_version(skill_id, version)`. If such a row exists, FK creation aborts the migration mid-way.
2. **Slug rewrite can violate the new unique indexes.** `_normalize_existing_files` rewrites `slug` â†’ `aai-<slug>` for `aai_cli` rows (lines 213-219); the partial unique indexes are created afterwards (lines 425-438). A pre-existing custom skill already slugged `aai-jira` in the same org â€” or two `aai_cli` rows normalising onto the same value â€” breaks index creation.

Both are cheap to pre-flight with an explicit `SELECT` that fails loudly, the way `_choose_entry_path` already does for ambiguous entry files. `test_skill_library_migration_backfill.py` covers the happy paths but not these.

---

### Non-blocking

| Where | Issue |
|---|---|
| `skills/service.py:801-815` (org) and `:545-557` (agent) | `apply_source_update` with **no draft** publishes the source version verbatim as the fork's next version â€” the fork's local divergence is discarded, not merged. The confirm copy in `skill-detail-page.tsx:382` ("Copy and publish the latest source versionâ€¦ Existing pins remain unchanged") never says the user's customisations are about to be overwritten. |
| `agent-skills-settings.tsx:76-84` | The "Credentials configured" badge renders whenever `requiredProviders.length > 0`, regardless of whether the agent actually holds those secrets. It replaced the previous, accurate provider list â€” so the display went from correct to actively misleading. |
| `skills/service.py:351`, `:233` | `is_assigned_to_agent` is hardcoded `False` for platform and agent scopes. It's `required` in the UI Zod schema (`skills/schemas.ts:39`) but read nowhere â€” compute it or drop the field. |
| `update_skill` / `update_agent_skill` / `update_platform_skill` | Read-then-`save` renames have no `IntegrityError` â†’ 409 translation, so a concurrent rename 500s. `_save_new_skill` already has exactly this translation; renames don't use it. |
| `agents/models.py:464` + siblings | The template-skill unique constraint moved from `(template_id, skill_id)` to `(template_id, skill_id, skill_version)`, which now permits one template to require the *same* skill at two versions. Both resolve to the same `root_dir` at mount time â€” suggest keeping the 2-column uniqueness. |
| `use-skill-mutations.ts:28-31` | `useCreateSkill`'s doc comment says "published as v1 immediately" â€” no longer true; create only opens a draft. |
| `skill-detail-page.tsx` | A freshly created (never-published) skill lands on the "Published v0 â€” This published version is read-only" tab with no Edit/Publish button until the user discovers the "Draft" toggle. `SkillNewPage:56` could route with `?edit=1`, which the page already supports. |

---

### AC gap

**"Select an exact published version â€¦ when requiring it from a Template"** is backend-only. `required_skill_versions` exists on `TemplateCreate` / `TemplateUpdate` / `PlatformTemplateDraft*`, but `grep -rn "requiredSkillVersions" ui/src` returns **zero hits** â€” template requirements always pin to latest from the app. Combined with #2, the API surface for it is also broken for mixed standalone+group edits.

Everything else in the AC looks implemented and covered: scope visibility, draft/publish/immutable versions, single root `SKILL.md` (enforced in `files.py:67-97`), version history and agent pin selection, non-repinning publish, fork + apply-update, the four delete guards, and runtime materialisation via `get_files_for_skill_versions`.

PR #144, opened 2026-08-26
AF-273 Communication diagnostics and recovery

kalibetre:
## Summary

AF-273 adds connection-scoped communication diagnostics, an append-only operational journal, delivery lifecycle visibility, and safe recovery controls for Slack, Telegram, Discord, and Microsoft Teams.

The implementation keeps the page-level experience condensed and incident-oriented: connection health is summarized, recent failures expand into safe details, and delivery transitions remain the primary activity explorer. Raw provider-observation and policy-admission events remain available through the connection journal API rather than being promoted to another page-level table.

### API and persistence

- Added the content-free `communication_operation_journal` model and migrations, including bounded retention.
- Added typed journal stages and inbound policy dispositions (`accepted`, `bot_ignored`, `mention_required`, `user_denied`, `channel_denied`, and `malformed_payload`).
- Added server-backed journal filtering, pagination, chronological per-delivery lifecycle reads, live delivery status enrichment, and Agent-scoped authorization predicates.
- Added connection diagnostics for provider health, reconnects, delivery counts, queue depth and age, latency, success rate, incidents, and recent failures.
- Added structured, redacted communication error details with safe category, operation, provider code, HTTP status, retryability, retry-after, and request ID fields.
- Added typed reconnect, retry, recovery, audit-event, and low-cardinality Prometheus metric paths.
- Preserved outbound ordering and stable idempotency identity across retries.

### UI

- Added the Communication Connection detail page and diagnostics surface.
- Added condensed health signals, incident grouping, safe failure-detail expansion, time-window filtering, paginated journal browsing, delivery status/direction badges, and per-delivery lifecycle drill-down.
- Added guarded reconnect and outbound dead-letter retry controls.
- Added the shadcn checkbox primitive used by the diagnostics filters.


dominykas-aai-labs:
## AF-273 review — Communication Operations (diagnostics & recovery)

Reviewed `ddd85595..HEAD` (73 files, ~7.6k insertions) — everything on the branch since the staging merge base.

The shape is right and the security-sensitive parts are done well: error redaction, metric label discipline, and the read/update permission split are all solid. Four things need resolving before merge; the rest are cleanups.

---

### 🔴 Blocking

**2. Domain Event flood during a connection failure loop**

`supervisor.py:104-167` retries on a flat `sleep(2)`; each pass calls `record_health(CONNECTING)` then `record_health(ERROR)`, and `repository.py:132-174` emits `communication.connection.health.changed` on *every* status change. A connection with bad credentials produces ~1 Domain Event/second into the outbox and the security-audit projection, plus ~86k journal rows/day.

The ticket is explicit: *"Repeated connection attempts, intermediate retries, and detailed timing belong in the operational journal and metrics rather than the Domain Event stream."*

Needs exponential backoff with a cap, and event emission debounced to entry/exit of a failure state.

**3. `end_to_end_health` reports "degraded" for normal in-flight traffic**

`operations.py:934-950`:

```python
if delivery_counts.pending or delivery_counts.processing:
    return "degraded"
```

`delivery_counts` covers deliveries created in the window (default 24h), so one message currently being processed flips a healthy Connection to amber. A single `UNAVAILABLE` delivery (agent was stopped when a message arrived) keeps it degraded for the whole window even after the agent is running again. This drives the headline card at `communication-connection-diagnostics.tsx:301-306`, so operators will see permanent amber on any busy Connection.

`oldest_pending_delivery_age_seconds` already exists — age in-flight work into degraded past a threshold instead.

**4. Two UI acceptance criteria are unmet**

> "The connection UI exposes a diagnostics view with a pipeline summary and the latest 50 transitions."

- `pipeline` is declared in `schemas.ts:85` and returned by the API but **never rendered anywhere**. There is no pipeline summary in the diagnostics component.
- `latestTransitions` (`schemas.ts:125`) is likewise never rendered. The view embeds a paginated journal filtered to `kind="delivery"` at page size 20 instead.
- Connection-scoped transitions (`kind="connection"`) aren't surfaced at all — `ui/tests/e2e/agent-detail-page.spec.ts:535` actively asserts that request is never made, so the backend `kind="connection"` branch (`operations.py:798`) is dead from the UI's perspective.

Either render these, or get the AC amended.

---

### 🟡 Medium

**5. The pipeline funnel can never show drop-off.** `gateway_service.py:202-214` writes `POLICY_ADMITTED` for every observed event regardless of disposition, so `provider_observed == policy_admitted` always. Write it only for `ACCEPTED`, or add a distinct rejected stage.

**6. Slack maps every non-message event to `BOT_IGNORED`.** `plugins/slack.py:273` — `reaction_added`, `member_joined_channel`, `message_changed`, `message_deleted` and every subtype all become "bot ignored". That's a specific operator-facing meaning per the ticket, and it pollutes both the `policy_dispositions` metric and diagnostics.

**7. A journal write failure rejects real provider messages.** `gateway_service.py:355-366` raises `CommunicationJournalWriteError` before any delivery exists, so a problem in the diagnostics table drops/500s genuine inbound messages. There's a test pinning this, so it's deliberate — but `PROVIDER_OBSERVED` is observability and shouldn't be able to take down ingress. Worth an explicit decision either way.

**8. `duration_ms` on connection-level rows is measured against unrelated rows.** `operations.py:988-1009` — for `delivery_id IS NULL` entries, `_duration_from_previous` picks the previous delivery-less row of *any* stage, including `PROVIDER_OBSERVED`/`POLICY_ADMITTED`. So duration on `CONNECTION_CONNECTED`/`CONNECTION_ERROR` is "time since last provider event", not connect time. (`_build_connection_history` already excludes admission rows — this should too.) Secondary: it adds a `SELECT` per journal insert, so each observed provider event costs 4 statements across two transactions.

---

<details>
<summary><b>🟢 Low / quality (9 items)</b></summary>

- **Optional injected deps hide wiring failures.** `operations: … | None = None` on five classes, plus `self.operations or getattr(self.repository, "operations", None)` in `service.py:192,251`. If injection regresses, journaling silently stops and diagnostics 503s with no signal. Make them required.
- `operations.py:249` — `stage_event` runs `select(EventDelivery.id)` and returns a list every caller discards.
- `events/catalog.py` now imports `CommunicationErrorDetails` from the communications domain — the generic event catalog depending on one domain's models is a layering inversion.
- `gateway_service.py:216-219` — `if not envelopes: return []` immediately followed by `if envelopes:`. Dead branch.
- `operations.py:838-841` — the journal page selects `CommunicationDelivery.direction, .status, CommunicationDelivery`; the entity already carries both columns.
- `operations.py:970-982` — `prune_journal` issues one unbounded `DELETE` every 5 minutes. Batch it before the table grows.
- `repository.py:196` — `request_reconnect` bumps `revision`, the Connection's optimistic-concurrency token. That's how the supervisor notices the restart, but an open edit form will 409 after anyone clicks Reconnect. Consider a separate restart-generation column.
- Function-local metric imports (`repository.py:266`, `delivery_repository.py:729`, `gateway_service.py:370`) — presumably cycle-dodging; a small metrics seam would read better.
- `find_journal_page` shadows the `direction` parameter with the comprehension variable. Harmless (comprehension scope) but confusing.

</details>

<details>
<summary><b>🧪 Test gaps</b></summary>

- No test that a user with **no** Agent access (or another tenant) gets 404 on `/summary` and `/journal`. `test_diagnostics_read_does_not_grant_connection_recovery_permission` covers viewer-vs-update; cross-tenant concealment is only tested for `/app-package`.
- No UI test for the "Retry delivery" control (backend idempotent retry is well covered).
- Nothing pins event volume during a reconnect storm — see finding 2.

</details>

---

### ✅ Done well

- **Error redaction** (`error_details.py`): category-derived codes, an allowlist of safe summaries, bounded/validated provider code and request id, and `_contains_sensitive` on every path into persistence. `test_connection_and_journal_reads_redact_legacy_error_details` covers legacy rows.
- **Metric cardinality**, with `test_communication_metrics_have_no_resource_identity_labels` enforcing it.
- **Retry preserves identity**: same delivery row, message row and idempotency key; `_BLOCKING_OUTBOUND_STATUSES` keeps conversation ordering intact while a dead-lettered row sits ahead in the stream.
- **Typed dispositions** across all four plugins, with `InboundAdmissionResult` kept sequence-compatible for existing call sites.
- **Journal schema**: no FKs by design (survives retirement), bounded configurable retention.

---

### Suggested order

1. Supervisor resilience (1) — this is the one that can silently disable all platform ingress in production.
2. Backoff + event debounce (2).
3. `end_to_end_health` semantics (3).
4. Pipeline summary + latest-50 transitions in the UI (4), and make `POLICY_ADMITTED` accepted-only so the funnel means something (5).
5. Slack disposition mapping (6), duration scoping (8), then the low-severity list.
6. Add the missing authorization and retry-UI tests.

PR #149, opened 2026-09-01
AF-279 Web Chat UI (Experimental)

kalibetre:


dominykas-aai-labs:
### Blockers

#### 1. ~40 concurrent agents per communications replica, then the gateway starves

`stream_runtime_control` is a **sync** generator behind `StreamingResponse`. Starlette drives it via `iterate_in_threadpool`, which takes an anyio worker-thread token per `next()` — and each `next()` blocks up to 15s inside `signals.wait` → `XREAD block=15000`. Token occupancy is effectively 100% per open stream.

anyio's default is `CapacityLimiter(40)`, nothing in the repo raises it, and `helm/agentbarn-api/values.yaml:145` sets `communications.replicaCount: 1`. Every route in `communications_app` is a sync `def`, so they all draw on the same 40 tokens.

At ~40 running agents the control streams hold the pool permanently and `claim` / `replies` / `complete` only get served in the gaps between a stream's yields. This is qualitatively new — v1 held no persistent connections.

`api/domains/communications/gateway_routes.py:43`, `api/domains/communications/gateway_service.py:98`

**Fix:** make the control stream an `async def` generator over an async Redis client (or `anyio.to_thread` with an explicit large limiter), and raise/verify the thread ceiling. Same applies to `api/domains/web_chat/routes.py:102`, though browser streams are fewer.

#### 2. Redis outage stops all v2 message processing — no fallback on the consume side

The publish side is defensive (`_publish_signal` swallows `RedisError`), but the consume side isn't: `CommunicationSignalBus.wait` catches only `RedisTimeoutError` and `latest_cursor` catches nothing. A connection error kills the SSE response; the adapter reconnects with backoff and fails again. A v2 pod has **no claim polling at all** — `DeliveryWorker` only drains when woken.

So a Redis blip degrades from "higher latency" (v1 behaviour) to "no agent processes any message until Redis returns."

**Fix:** give `DeliveryWorker` a slow safety-net claim (e.g. `self._wake.wait(timeout=30)`) so Redis loss costs latency, not delivery.

#### 3. AC #4 is not met once a delivery reaches PROCESSING

> *"the composer returns to 'ready' state within a few seconds"*

`useWebChat` derives `isAwaitingReply` from the last message's `deliveryStatus` being `PENDING`/`PROCESSING`. For a PROCESSING delivery, `request_cancel` sets `cancel_requested_at` but deliberately leaves the status at `PROCESSING` — it only becomes `CANCELLED` when the runtime posts `/complete`, i.e. after the soft-cancelled LLM call finishes. That can be minutes.

The only stop test, `test_stop_generation_cancels_the_active_thread_delivery_idempotently`, sends into an agent that never claims — so it exercises the PENDING path and asserts `CANCELLED` immediately. It doesn't cover the case the AC describes.

**Fix:** surface `cancel_requested_at` in `WebChatMessageRead` (a `cancelRequested` flag, or a `CANCELLING` display status) and clear `isAwaitingReply` on it. The durable status can stay PROCESSING until completion.

---

### High

**4. Long threads freeze — oldest 500 messages, no new ones.** `api/domains/web_chat/repository.py:28` — `list_thread_messages` is `ORDER BY id ASC LIMIT 500`. Past 500 messages a thread shows only its oldest 500 forever, and since `stream_updates` uses the same call, live updates stop too. Order DESC + limit then reverse, or page from the tail.

**5. `stream_updates` does full re-authorization and a full thread re-read per signal.** `api/domains/web_chat/service.py:185` calls `self.list_messages(...)` inside the wait loop, re-running `authorization.require_action`, `_get_or_create_connection`, the message query and `delivery_statuses_for_messages` on *every* agent-scoped signal, for every open browser stream. `emitted` also grows without bound. `list_messages` already accepts `after_id` (UUIDv7, documented as a valid cursor) but the stream always passes `None`. Hoist auth/connection lookup out of the loop and use `after_id` for appends, keeping a small recent-id window for status-change diffs.

**6. Adapter hot-reconnects on a clean stream close.** In `communications-runtime-adapter.py`'s `main()`, if `consume_control_stream` returns normally (server closed the response cleanly — rolling restart, LB idle cut) it sets `reconnect_delay = 1` and immediately re-enters with **no sleep**. Only the exception path sleeps.

**7. Cancel is dropped if it arrives in the claim→`begin()` window.** `run_delivery` calls `IN_FLIGHT.begin(delivery_id, …)` *after* the claim returns. A `delivery_cancelled` for that id in between hits `request_cancel`'s `self._delivery_id != delivery_id` guard and is silently discarded, so the runtime call is never marked. Durability still holds — the reply POST gets a 409, the generic `except` converts it to a failed completion, `_apply_completion` marks it CANCELLED — but the user pays full LLM latency and the outcome rides on an error path rather than the intended one. Keep a small set of "cancel requested for an id not yet begun" and have `begin()` consult it.

---

### Medium

**8. `per-file-target-version` doesn't cover the adapter.** `api/pyproject.toml:54` guards `domains/agents/scripts/openclaw/**` and `hermes/**`, with a comment explaining these run on the pod's 3.12/3.13 interpreter and that ruff would otherwise rewrite `except (A, B):` into the 3.14-only form. But `communications-runtime-adapter.py` sits one level up at `domains/agents/scripts/`, is read by *both* builders (`hermes.py:18`, `openclaw.py:24`), and matches neither glob. It parses on 3.13 today — it's one `ruff format` away from shipping a `SyntaxError` into every agent pod. Add `"domains/agents/scripts/communications-runtime-adapter.py" = "py312"`.

**9. The supervisor still leases and parks a task per Web Chat connection.** `_maintain` decrypts credentials, writes CONNECTING then CONNECTED, and parks on `asyncio.Event().wait()` — for a plugin with nothing to supervise, once per agent. Cleaner to filter non-supervised plugins out of the enumeration around `api/domains/communications/supervisor.py:93-120` so they never take an ingress lease.

**10. `SUPERVISED_INGRESS` fails open, untested.** A future plugin that forgets the flag silently becomes unsupervised and reports CONNECTED — the opposite of the loud `INGRESS_NOT_SUPERVISED` this replaces. There's no test anywhere for the new supervisor branch, despite it being the shared-infra change flagged as the main regression risk. Add a test that each shipped plugin either declares `SUPERVISED_INGRESS` or overrides `run_ingress`.

**11. AC #7's fix is untested.** The `MESSAGE_CHANGED` publish in `complete_runtime_delivery` is the entire fix for the stuck "is working" state, and nothing asserts it (`MESSAGE_CHANGED` has zero hits in `api/tests/`). Nothing covers `stream_updates` either.

**12. SSE bypasses token refresh.** `ui/src/features/agents/hooks/use-sse-stream.ts:64` reads `useAuthStore.getState().authToken` with raw `fetch`, so it never passes through `AuthInterceptor`. When the access token expires mid-session the reconnect gets 401, falls into `!response.ok`, and loops at 15s backoff with the same stale token indefinitely — the panel reads "Updates paused" until a page reload.

**13. Reads provision rows.** `_get_or_create_connection` is called from `list_messages`, `list_threads` and `stream_updates`, which require only `ACTIVITY_READ` — a GET creates a `communication_connection`. The unique-index race is handled, but a read endpoint doing a write deserves a deliberate decision.

**14. `thread_id` path params are unvalidated.** `rename`/`delete`/`stop` take a bare `str` while `web_chat_thread.thread_id` is `max_length=128` — a long value from `?thread=` yields a DB error rather than a 422. `POST /messages` does constrain it. Relatedly, the UI `encodeURIComponent`s the thread id for rename and delete but not for the stop URL.

**15. `list_threads` is unbounded.** Both `DISTINCT ON` queries select full `AgentChatMessage` rows for every thread, then sort and slice to `limit` in Python.

**16. `AGENT_RUNTIME_KIND` is dead.** Added to both builders (and asserted in one test) but read nowhere.

**17. `stop_generation`'s return value is discarded** by the route, which always answers 204 — the UI can't distinguish "cancelled" from "nothing in flight".

---

### Low

- `find_active_inbound_delivery` cancels a single delivery; a second message queued in the same thread still runs after Stop, so the UI flips ready→working again.
- A reply enqueued microseconds before Stop still lands, while its inbound turn is recorded CANCELLED.
- `inputRef` in `ThreadListItem` (`ui/src/features/agents/components/chat-tab.tsx:115`) is assigned but never used — `autoFocus` does the work.
- `useWebChatThreads` returns a fresh object and fresh callbacks each render while `useWebChat` memoizes. Harmless today, inconsistent.

dominykas-aai-labs (approved):

PR #150, opened 2026-09-02
Af 265 fix agent analytics staging

dominykas-aai-labs:


api/domains/agents/repository.py
```diff
@@ -209,13 +210,13 @@ def _stats_predicates(
...
-                    col(CommunicationConnection.retired_at).is_(None),
-                )
-            )
+            connection_predicates = [
+                col(CommunicationConnection.agent_id) == col(Agent.id),
+                col(CommunicationConnection.platform_key) == platform.value,
+            ]
+            if not include_retired_connections:
+                connection_predicates.append(col(CommunicationConnection.retired_at).is_(None))
+            predicates.append(exists().where(*connection_predicates))
```
kalibetre, line 221: **[Standards · Judgement] Duplicated Code** — This `exists()` predicate shape (agent/connection join, `platform_key` match, optional `retired_at` exclusion) is now written independently in three repositories: here, `conversations/repository.py:205-210`, and `tool_calls/repository.py:180-185`. Only this one is parameterized with `include_retired_connections`; the other two disagree on retired-connection handling with no shared source of truth.

Suggest extracting a shared predicate builder so the three can't drift further.

api/domains/agents/repository.py
```diff
@@ -197,6 +197,7 @@ def _stats_predicates(
         agent_id: UUID | None,
         created_by_user_id: UUID | None,
         platform: CommunicationPlatform | None,
+        include_retired_connections: bool = False,
```
kalibetre, line 202: **[Standards · Judgement] Mysterious Name / thin comment + minor Data Clump** — `include_retired_connections` has no docstring note on when callers should pass `True`; the point-in-time-vs-historical rule that justifies it currently only lives in `docs/features/platform-administration/CHANGELOG.md`, not at this call site. Also a minor boolean-flag-param smell, though low-stakes given there's only one alternate caller today.

api/domains/conversations/repository.py
```diff
@@ -195,26 +195,28 @@ def daily_active_agent_ids_since(
...
+        if agent_id is not None:
+            message_predicates.append(col(AgentChatMessage.agent_id) == agent_id)
         if platform is not None:
-            agent_predicates.append(
+            message_predicates.append(
                 sa.exists().where(
                     col(CommunicationConnection.id) == col(AgentChatMessage.connection_id),
                     col(CommunicationConnection.platform_key) == platform.value,
                 )
             )
```
kalibetre, line 210: **[Standards · Judgement] Duplicated Code** — Same predicate shape as `agents/repository.py:213-219` and `tool_calls/repository.py:180-185`, but this instance has no retired-connection handling at all (unconditionally includes retired connections). See the comment on `agents/repository.py:213-219` for the suggested fix.

api/domains/tool_calls/repository.py
```diff
@@ -176,18 +176,17 @@ def daily_active_agent_ids_since(
...
-                            col(CommunicationConnection.platform_key) == platform.value,
-                            col(CommunicationConnection.retired_at).is_(None),
-                        )
+            if platform is not None:
+                query = query.where(
+                    sa.exists().where(
+                        col(CommunicationConnection.agent_id) == col(ToolCall.agent_id),
+                        col(CommunicationConnection.platform_key) == platform.value,
                     )
+                )
```
kalibetre, line 185: **[Standards · Judgement] Duplicated Code + [Spec · Suspicious Implementation]** — Same predicate shape as the other two repos, but the previous `retired_at.is_(None)` check was dropped here entirely with no comment explaining why, while `agents/repository.py` treats this as a deliberate opt-in (`include_retired_connections=True`). Given AF-265 AC1 ("show exact ... tool call counts"), worth confirming this is intentional rather than an oversight — as written it silently includes retired connections in tool-call stats.

api/tests/integration/test_platform_stats.py
```diff
@@ -750,3 +785,231 @@ def test_agent_stats_reports_zero_on_an_empty_platform():
         assert_that(body["total"], equal_to(0))
         assert_that(body["running"], equal_to(0))
         assert_that(len(body["observed_at"]), greater_than(0))
+
+
```
kalibetre, line 789: **[Standards · Judgement] Divergent Change** — This file is touched for two different reasons in one commit: rewriting seeding helpers to route through the Communications Gateway (mechanical/harness change), and adding this new ~230-line contract-assertion suite (product invariants) starting here. Likely fine bundled under one ticket — flagging as a judgement call only, not a blocker.

ui/src/features/platform-stats/components/platform-stats-panel.tsx
```diff
@@ -248,17 +262,17 @@ export function PlatformStatsPanel() {
       >
         <StatTile
           label="Messages"
```
kalibetre, line 264: **[Spec · Missing Requirement]** AF-265 AC1: "The message count on the staging platform show exact message and **tool call counts** of agents in staging based on the filter." This panel only ever surfaces Messages/Received/Sent/Active-agents — no tool-call count tile. "Active agents" counts distinct agents, not tool calls, so AC1 isn't satisfied as written.

ui/src/features/platform-stats/presets.ts
```diff
@@ -1,132 +0,0 @@
-import type { StatsRange } from "./schemas";
```
kalibetre, line 1: **[Spec · Missing Requirement]** AF-265 AC2 asks to remove 6 named presets ("today, yesterday, last 7 days, last 30 days, last week, last month"), implying the rest stay. This deletes the entire file — every preset, not just the 6 named ones — replaced by a raw `DateRangePicker`. The CHANGELOG's claim that "Last hour, Last 6/12 hours, This week, This month, This year, Custom range" remain doesn't match what shipped. Worth confirming the broader removal was intentional, and fixing the CHANGELOG either way.

kalibetre (commented):
**No hard standards violations.** The judgement calls center on the same `exists()` platform predicate now duplicated (and drifting) across `agents/`, `conversations/`, and `tool_calls/repository.py` — worth a shared helper.

**Spec gaps are the bigger concern:** AC1 ("exact message **and tool call** counts") isn't implemented — no tool-call count is surfaced in the UI. AC2 (remove 6 named presets) shipped as "remove *all* presets," which the CHANGELOG doesn't accurately describe. There's also meaningful scope creep beyond the two ACs: a new message-direction filter, a new `TEAMS` platform, and a new URL-state-persistence hook/`Suspense` wrapper — none requested by this ticket.

Inline comments below for specifics.

ui/src/features/platform-stats/schemas.ts
```diff
@@ -53,7 +53,9 @@ export const PlatformAgentStatsSchema = z.object({
   series: z.array(PlatformAgentSeriesPointSchema),
 });
 
-export const AgentPlatformSchema = z.enum(["slack", "telegram", "discord"]);
+export const AgentPlatformSchema = z.enum(["slack", "teams", "telegram", "discord"]);
+
+export const MessageDirectionSchema = z.enum(["all", "inbound", "outbound"]);
```
kalibetre, line 58: **[Spec · Scope Creep]** New message-direction filter (`all`/`inbound`/`outbound`) isn't requested by either AC1 or AC2. Consider a separate PR/ticket unless it's a hard prerequisite for the AC1 fix.

api/domains/communications/models.py
```diff
@@ -27,6 +27,7 @@ class CommunicationPlatform(str, enum.Enum):
     """Stable keys for the communication plugins shipped with Agent Barn."""
 
     SLACK = "slack"
+    TEAMS = "teams"
```
kalibetre, line 30: **[Spec · Scope Creep]** Adding `TEAMS` here reads like an unrelated plugin-registration fix folded into AF-265. Worth splitting into its own PR unless it's actually load-bearing for this ticket.

ui/src/features/platform-stats/hooks/use-stats-url-state.ts
```diff
@@ -0,0 +1,95 @@
+"use client";
```
kalibetre, line 1: **[Spec · Scope Creep]** This new 95-line hook (URL-state persistence for from/to/org/platform/direction) isn't requested by either AC. The CHANGELOG's "no API contract moved" framing undersells the new surface area it adds. Also drives the `Suspense` wrapper added in `ui/src/app/dashboard/platform/page.tsx:90-92`.

ui/src/app/dashboard/platform/page.tsx
```diff
@@ -86,7 +87,9 @@ export default function PlatformPage() {
           ))}
         </div>
 
-        <PlatformStatsPanel />
+        <Suspense fallback={null}>
+          <PlatformStatsPanel />
+        </Suspense>
```
kalibetre, line 98: **[Spec · Scope Creep]** This `Suspense` boundary is only needed because of the new `useSearchParams`-based URL-state hook (`use-stats-url-state.ts`) — a downstream effect of that scope-creep addition, not something either AC calls for.

kalibetre (approved):

PR #167, opened 2026-09-06
Fix Agent DM amnesia, add verbose-mode progress visibility, harden delivery reclaim (AF-386)

kalibetre:
This PR bundles four related fixes to the Hermes ↔ Communications runtime path, triggered by a report that a Hermes agent would intermittently forget earlier turns in the same Telegram DM thread.

1. Hermes DM/thread history continuity (root cause fix)
Hermes's /v1/runs API treated session_id as a persistence identity only — it never reloaded prior messages into the model's context on later turns, so the agent appeared to forget the conversation despite history still existing in Hermes's SQLite store. Every delivery now sends resume_session: true, and hermes-base/patch-run-session-history.py patches the pinned Hermes image to actually resume from stored history (compaction-aware, tool-metadata-preserving), failing the build if the upstream source drifts.

2. Async /v1/runs adapter rewrite + verbose_mode setting
Supporting per-turn resume required replacing the adapter's single blocking chat-completions call with Hermes's async /v1/runs + SSE run API. This also unlocked a new optional verbose_mode Agent setting (Hermes-only, UI toggle) that relays mid-turn tool/subagent progress as clear chat messages instead of silence. Rejected for OpenClaw agents, since OpenClaw has no external HTTP channel for these events yet (tracked as a follow-up).

3. Communications delivery reclaim hardening
Inbound deliveries whose processing lease expired without a /complete call previously stayed stuck PROCESSING forever, blocking the rest of that conversation's queue. They're now reclaimed with backoff and dead-lettered after repeated failures, scoped per-agent to avoid a cross-tenant sweep.

4. Telegram message chunking
Outbound messages over Telegram's 4096-character limit previously failed outright; they're now split and sent as multiple messages.


dominykas-aai-labs:
### 🔴 1. Long Hermes turns churn through the reclaim and dead-letter mid-run

The inbound lease is a hard-coded 120s (`api/domains/communications/delivery_repository.py:167`), there is no renewal or heartbeat anywhere in the codebase, and `_drain_run` holds the SSE stream for up to 900s (`communications-runtime-adapter.py:307`).

Any turn longer than 120s now:
1. gets reclaimed as `LEASE_EXPIRED` (attempt++), backoff `2**attempt` = 2–16s
2. is re-claimed, hits the `_ACTIVE_RUNS` guard, which returns without completing
3. repeats until `attempt_count >= 5` → `DEAD_LETTERED` (~10 min wall clock)

When the run finally finishes, `/complete` 404s because `complete_runtime_delivery` requires `PROCESSING` (`delivery_repository.py:530`). The adapter's `except` handler calls `complete_delivery` again, gets another 404, and the exception escapes the daemon thread. The reply still reaches the user (`enqueue_runtime_reply` doesn't check status), but the delivery is recorded as a failure, and every long turn emits up to four spurious `LEASE_EXPIRED` journal rows plus repeated `CLAIMED` processing-feedback pings to the provider.

Not hypothetical — long agentic turns and indefinitely-blocking human approvals are exactly what the async run API was introduced to support.

**Suggested fix:** add a lease-renewal endpoint the drain thread heartbeats every ~60s (matching the existing `claim_ingress_lease` pattern in `repository.py:292`), or pass a `lease_seconds` from `claim_runtime_delivery` matched to the runtime timeout while keeping 120s for OpenClaw.

### 🔴 2. A second message in the same conversation is silently dropped

`communications-runtime-adapter.py:226-238` returns without replying or completing when `_ACTIVE_RUNS` holds a live thread for the session. The comment claims *"the live thread will reply/complete this same delivery_id"* — true only when the server re-hands the **same** delivery.

It doesn't. Once the first delivery's lease expires it flips to `PENDING` with `available_at = now + backoff`; a message that arrived meanwhile has an **earlier** `available_at`, and the claim query orders by `available_at asc, created_at asc, id asc` (`delivery_repository.py:215-219`). So the *new* delivery is handed out first, is swallowed by the guard, never completed, and dead-letters after five lease cycles. The user's second message gets no answer and no error.

The existing test encodes this gap rather than closing it: `tests/unit/test_communications_runtime_adapter.py:258` passes `delivery-2` and asserts only that no second run starts — nothing asserts delivery-2 is ever replied to or completed.

**Suggested fix:** make the guard compare `delivery_id`. A reclaim of the same id stays a no-op; a different id gets handed to the running thread, or an immediate "still working on the previous message" reply + `complete(succeeded=True)`. Then assert that disposition in the test.

### 🔴 3. The ticket's headline fix has no test at any level

`hermes-base/patch-run-session-history.py` is guarded only by a single-occurrence anchor match plus `compile()`. `compile()` catches syntax errors only. The injected block references `_openai_error`, `previous_response_id`, `raw_input`, `self._ensure_session_db_async`, `db.resolve_resume_session_id`, `db.get_messages_as_conversation`, and assigns `conversation_history`. If upstream renames any of those — or recomputes `conversation_history` *after* the anchor — the build stays green and the original bug returns silently.

`api/runtime_tests/` now exists and runs against the real image in CI, but contains only skill discovery.

**Suggested fix:** add `api/runtime_tests/test_hermes_run_session_history.py` following the `test_hermes_skill_discovery.py` pattern — boot the patched image, POST two `/v1/runs` with the same `session_id` and `resume_session: true`, assert the second turn sees turn 1; assert `resume_session` + explicit history returns 400; assert a fresh session starts empty rather than erroring.

**Related:** the adapter sends `resume_session: true` unconditionally, and an unpatched image would ignore the unknown field and regress silently. The docs say roll them together, but nothing enforces it — a startup probe, or a version marker the patch writes and the adapter asserts, would make the coupling fail loudly.

---

### 🟡 Should fix before merge

**4. Verbose mode has no throttle.** Every `tool.started` becomes its own outbound delivery → its own provider message. A 30-tool turn is 30 Telegram messages; Telegram allows roughly 1/s and ~20/min per chat. Coalesce into a single edited message, or rate-cap the relay.

**5. Progress relaying isn't best-effort.** A failed `post_reply` inside `_drain_run` (`communications-runtime-adapter.py:339`) aborts the whole drain, so a transient failure on a *progress* message fails the turn while the Hermes run keeps going unattended. Wrap the verbose and approval relays in try/except.

**6. Lease-expiry dead-letters are invisible.** The reclaim path calls `_apply_completion` and `_stage_completion_journal` but not `_record_completion_metric`, and not the FAILED processing feedback that `complete_runtime_delivery` performs (`gateway_service.py:118`). A conversation that dies this way is missing from delivery metrics and the user just gets silence.

---

### 🔵 Worth noting, not blocking

**7.** `_PENDING_APPROVALS` / `_ACTIVE_RUNS` are process-local. The comment acknowledges this for approvals, but the blast radius is larger now that runs are long-lived: a pod restart mid-approval strands the Hermes run and the user's "once" starts a fresh one.

**8.** The Hermes main loop is now non-blocking, so a pod starts unbounded concurrent run threads across conversations (previously strictly one at a time). Each holds an SSE socket up to 900s. Consider a bounded worker pool, or document the cap.

**9.** Telegram chunking (`api/infrastructure/telegram/client.py`):
- A failure on chunk 2 retries the whole outbound delivery and re-sends chunk 1. Telegram has no native idempotency; the `:{index}` key only helps behind an egress proxy.
- The 4096 limit is UTF-16 code units, not Python characters — a 4096-char chunk containing emoji can still be rejected.
- `send_message` now returns the **last** chunk's message id where callers previously got the only one.
- `lstrip("\n")` at a split boundary drops deliberate blank-line spacing.

**10. Scope:** the branch also carries the Discord close-code diagnostics, the Hermes PVC root init container, and `skills.external_dirs`. All look self-consistent, but none are in this ticket — they arrived via staging merges. Worth confirming they're meant to ship here.

kalibetre:
Addressed in 17e86978:

1. Long Hermes turns now renew their authenticated inbound-delivery lease every 60 seconds through a new runtime protocol endpoint. A renewal cannot revive an already-expired claim.
2. The active-run guard now distinguishes a reclaim of the same delivery from a later delivery in the same session. The latter gets a durable, completed “still working on your previous message” reply; coverage asserts both the reply and completion.
4. Verbose progress relay is capped to one message per three seconds.
5. Progress and approval relay failures are now best-effort, so they cannot abort the run drain; final replies remain delivery-critical.
6. Lease reclaim now records delivery-outcome metrics, and newly dead-lettered reclaimed deliveries issue the normal FAILED provider feedback.

I also added repository/gateway/adapter regression coverage and updated the runtime contract documentation. `make check-api` and the focused 37-test communications suite pass.

kalibetre:
Clarification on item 3: we are intentionally handling this as reviewer manual verification, not adding a repository E2E test in this PR. A meaningful test must boot the real pinned Hermes image and drive actual `/v1/runs` behavior; that is substantially heavier than the adapter/repository regression suite and requires a runtime-capable environment.

Please manually verify against the built image: run two turns with the same `session_id` and `resume_session: true`, confirm turn two receives turn-one context, confirm `resume_session` plus explicit history returns 400, and confirm a fresh session has no inherited context. The patch still has its build-time anchor/compile guard, while this manual run validates the upstream runtime behavior end-to-end.

kalibetre:
Follow-up disposition for the remaining review notes:

- **7: restart during active run/approval:** accepted as a durability follow-up. Current behavior should explicitly tell the user that the previous run was interrupted and request a retry; a durable run/approval-state design can follow separately.
- **8: unbounded Hermes run threads:** accepted as a separate capacity-control PR. It needs a bounded worker/concurrency policy plus observable queue or rejection behavior.
- **9: Telegram chunking:** the contained correctness fixes are now included in 0949c654: chunks use Telegram’s UTF-16 code-unit limit and preserve newline/blank-line content exactly. Partial-send retry/idempotency and the logical-message return-ID contract remain separate design work.
- **10: staging-merged changes:** confirmed as intended branch content; no action required in this PR.

Telegram regression coverage passes (`api/tests/unit/test_telegram_client.py`, 14 tests).

dominykas-aai-labs (approved):

PR #172, opened 2026-09-07
fix(agents): serialize lifecycle mutations to stop concurrent starts desyncing credentials

### LINKED ISSUE #160: Cconcurrent Agent starts can desynchronize runtime and persisted credentials
## Summary

  Two concurrent start requests for the same Agent can both succeed.

  Each request generates new runtime credentials. Because the lifecycle status is
  checked before any cross-process serialization, both requests may proceed to
  runtime provisioning and persist different credentials.

  This can leave the Kubernetes Secret containing credentials from one request
  while the database contains credentials from the other.

  ## Expected behavior

  Only one lifecycle mutation should run for an Agent at a time. A competing
  request should receive HTTP 409 Conflict.

  ## Actual behavior

  Both concurrent start requests can return HTTP 200. The Agent is recorded as
  running, but its persisted credentials may not match the credentials mounted
  into the runtime.

  This can cause ingest or communication authentication failures even though the
  Agent appears healthy.

  ## Reproduction

  1. Create a stopped Agent.
  2. Send two start requests for that Agent concurrently.
  3. Allow both requests to reach runtime provisioning before either commits.
  4. Observe that both requests succeed.
  5. Compare the decrypted database ingest key with the key in the Kubernetes
     Secret.

  In a deterministic integration reproduction, both responses were HTTP 200 and
  the two keys differed.

  ## Root cause

  Agent lifecycle state was checked before Kubernetes operations, while the
  database row was locked only during the final write. That lock did not prevent
  both requests from independently provisioning the runtime.

  ## Impact

  Medium-to-high reliability impact:

  - The Agent may appear running but reject authenticated runtime communication.
  - It can be triggered through multiple tabs, users, direct API clients, or
    retries.
  - It affects one Agent at a time and is recoverable.
  - No authorization bypass or data loss has been identified..

## Proposed fix

  Serialize runtime-mutating operations per Agent across API processes. After
  acquiring the lock, reload the Agent and recheck authorization and lifecycle
  state. Return HTTP 409 Conflict when another lifecycle operation is already in
  progress.

  A regression test should verify that concurrent start requests cannot leave
  the runtime Secret and persisted credentials out of sync.

kalibetre:
## Summary

- Two concurrent start requests for the same Agent could both pass the status check before either committed, each generating its own runtime credentials and independently provisioning Kubernetes resources — the DB write was the only lock, and it ran after the runtime side effects, so it decided which credentials were persisted but not which were actually running in the Secret.
- Add `AgentRepository.lifecycle_lock`, a per-agent non-blocking Postgres advisory lock held for the full duration of start/stop, mirroring the existing advisory-lock pattern used for template-key allocation. A losing concurrent request now gets `409` immediately instead of racing through provisioning.
- Wire the lock into `service.py` around the start/stop lifecycle mutations.
- Fixes aai-labs/agent-barn#160 (AF-287).
- Unrelated migration fix bundled in: merge the two unmerged Alembic heads (`50400a4c8a00` and `d5a48a7fe943`) on staging so plain `alembic upgrade head` and the local test suite work again.

## Testing

- `api/tests/integration/test_agents.py` — new coverage for concurrent start/stop lifecycle locking (added in this branch).


api/domains/agents/service.py
sami-bre: `delete_agent` never takes the lock, so a delete landing mid-start tears down the k8s objects the start is still creating and leaves an orphaned Deployment plus a RUNNING row that's already soft-deleted.

sami-bre: Same thing here

api/domains/agents/service.py
```diff
@@ -1843,7 +1843,17 @@ def update_agent(self, agent_id: UUID, data: AgentUpdate, context: CurrentUserCo
...
+        actor = resolve_actor_identity(context, agent.organization_id)
+        with self.repository.lifecycle_lock(agent.id) as acquired:
+            if not acquired:
+                raise HTTPException(
+                    status_code=status.HTTP_409_CONFLICT,
+                    detail=f"Agent {agent_id} has a lifecycle operation already in progress",
+                )
+            # Re-read status now that we hold the lock: it may have changed since
+            # the caller's copy was loaded above.
+            current = self.repository.get_by_id(agent.id) or agent
```
sami-bre: This fails open in exactly the case the re-read exists to catch. `get_by_id` filters `deleted_at IS NULL`, so if the agent is soft-deleted between `require_action` and the lock, `or agent` hands back the stale pre-lock copy and the start proceeds: a full k8s workload (config map, secret, PVC, service, deployment) provisioned for a row that no longer exists, then `status=RUNNING` committed on it. Nothing in the UI points at that Deployment afterwards.

Every other `get_by_id` caller treats `None` as fatal — `ingest/service.py:25`, `communications/gateway_service.py:70`, `communications/processor.py:59`. Suggest raising 404 here instead of falling back. Same line in `stop_agent` at 2316.

sami-bre (changes requested):
Lock design looks right and the concurrency test is genuine — one blocker before merge, inline below.

Non-blocking, for a follow-up: `delete_agent` and `rebuild_running_agents_for_maintenance` mutate the same k8s objects but never take `lifecycle_lock`, so a delete landing mid-start still leaves an orphaned Deployment. Worth either pulling `delete_agent` into this PR or narrowing the `lifecycle_lock` docstring, which currently claims to serialize lifecycle mutations generally.

sami-bre (approved):
All looks good now

PR #180, opened 2026-09-08
feat(communications): opt-in admission of explicitly addressed bot messages on Slack

yeabtsega07:
## What

Adds an opt-in **Mentions from other bots** setting (`accept_bot_mentions`, default off) to Slack Connection settings.

- With it enabled, a message written by any Slack bot (including another Agent Barn Agent) is admitted only when it explicitly @mentions this Agent. Without a mention it is journaled as `mention_required`.
- The Agent's own messages stay `bot_ignored`, bot messages never inherit `start_only` thread ownership (so two Agents cannot keep each other awake on unmentioned thread replies), and admission fails closed when ingress did not capture the bot identity or when a legacy bot event carries only `bot_id` (a different id space from the bot user id, so a self-authored legacy event cannot be recognised).
- Default off preserves today's behaviour exactly: every bot-authored message is refused as `bot_ignored`.

## Scope, after review

This PR is **opt-in admission of explicitly addressed bot messages**, not full agent-to-agent collaboration. Trust decision, documented in the setting description and in `runtime-and-deployment.md`: with the toggle on, any Slack bot that shares a permitted channel can address the Agent, so the setting is off by default and intended for Agents working in controlled channels.

Deliberately out of scope here and proposed as follow-up design work owned by the Agent Barn team: peer discovery for Agents, a known-peer trust boundary (same Organization and workspace), sender provenance passed to the recipient runtime, exclusive active thread ownership for `start_only`, multi-recipient semantics, and bounded context transfer.

## Why

Today the Slack plugin drops every event carrying `bot_id` before mention checks run, so one Agent can never hear another. We want several Agents to hand work to each other inside one channel (an AAI Labs blogger pipeline: a scout agent opens a thread and @mentions a writer agent, which @mentions a reviewer agent). Agreed with the Agent Barn team to contribute this as a PR (internal refs BAW-247 / BAW-257).

The Slack manifest already subscribes to `message.channels`, so no Slack app change is needed; operators flip the setting per Connection.

## Changes

- `api/domains/communications/plugins/slack.py`: new `SlackSettings.accept_bot_mentions`; bot-authored events are admitted only on an explicit mention of this Agent, decided inside `normalize_inbound` so no later thread policy can widen it. Legacy bot events without a `user` fail closed.
- `api/tests/unit/test_communications_plugins.py`: seven tests covering the default, mention/no-mention/other-agent-mention, own echo, `start_only` threads, missing bot identity, legacy `bot_id`-only events failing closed (toggle on and off), and Slack's documented modern bot-message shape (plain `message` event with `bot_id` + `bot_profile`, no subtype; see [docs.slack.dev](https://docs.slack.dev/reference/events/message/bot_message)).
- Docs: `docs/architecture/runtime-and-deployment.md` (mention gating), `docs/features/agents.md` (shared-room behaviour), `docs/features/communications/CHANGELOG.md` (current state + dated entry).

No migration (new optional key in the plugin-owned settings JSON). No UI change: the Connection form is schema-driven and already renders boolean properties.

## Checks run

- `make check-api` (ruff check, ruff format --check, ty) — passes
- `make check-migrations` — passes
- `make test-api` (full API unit + integration suite) — 1914 passed
- Slack plugin tests alone (`-k slack`) — 23 passed

Not run: UI suites (no UI files changed), `make test-api-k8s`.

## Notes for review

- Disposition choice: an unmentioned bot message with the setting on is reported as `mention_required` rather than `bot_ignored`, matching how an unmentioned human message is reported, so the journal distinguishes "bots are off" from "bot spoke but did not address me". Happy to change if you prefer a dedicated disposition.
- Ping-pong between two Agents that keep @mentioning each other is a prompt/template concern, same as for humans; this change only governs admission.


kalibetre (changes requested):
## Agent-to-Agent Slack Collaboration

**Request changes.** This is a solid receiving-side foundation, but it does not yet provide a complete, natural, or safe agent handoff workflow.

The change correctly:

- preserves the default of ignoring bot-authored Slack messages;
- requires an explicit target mention when the opt-in is enabled;
- prevents bot messages from inheriting `start_only` thread ownership;
- handles the modern Slack bot-message event shape;
- keeps the normal self-message path ignored.

However, “admit a bot message that already contains `<@TARGET_ID>`” is not the same as enabling agents to collaborate.

## 1. Agents cannot discover or produce peer mentions

The core intended workflow is:

```text
Agent A receives work
→ Agent A decides Agent B should continue it
→ Agent A identifies Agent B
→ Agent A posts a real Slack mention for B
→ Agent B receives the handoff
```

The current implementation only covers the final step. A running Agent has no reliable way to learn Agent B’s Slack bot user ID or generate the required `<@USER_ID>` token. The existing Slack directory path is operator-facing, and runtime input/output is still raw text.

Please add a narrow, tenant-scoped peer-resolution capability that exposes eligible peers in the active Slack workspace:

```text
Agent display name → Slack bot user ID → rendered Slack mention token
Researcher         → U123ABC          → <@U123ABC>
Writer             → U456DEF          → <@U456DEF>
```

This can be injected as ephemeral runtime context or exposed through a narrow runtime tool. It should not expose credentials or general Slack-directory access.

## 2. Define which bots are trusted handoff senders

`accept_agent_mentions` currently admits a direct mention from any Slack bot. That includes external Slack apps, not only Agent Barn Agents.

Please make the trust boundary explicit. The safer default is to admit only known peer Agent Barn Connections satisfying:

```text
recipient opted in
+ same organization
+ same Slack workspace
+ known eligible peer Connection
+ exact recipient mention
```

If arbitrary external bots are intentionally allowed, document that as a security decision: those bots can submit instructions to an opted-in Agent whenever they share a permitted channel.

## 3. Pass structured handoff context to the recipient

The normalized inbound envelope retains sender identity, but the runtime receives only the message text. The recipient therefore cannot reliably tell whether it was contacted by a human, an external bot, or another Agent Barn Agent.

A receiving Agent should get safe metadata such as:

```text
Source: Agent Barn peer
Sender: Researcher
Sender Slack identity: U123ABC
Target: Writer
```

This makes collaboration intelligible and allows the recipient to reason about provenance without trusting text supplied by the sender.

## 4. Define thread ownership for `start_only`

A Slack thread is a shared presentation surface. It must not be treated as a shared Agent execution session.

The current architecture keeps runtime sessions separate per Connection, which is good: Agent B does not automatically inherit Agent A’s private session history simply because both reply in the same Slack thread.

But after a handoff, both Agents may become historical owners of the same thread:

```text
A starts / owns thread
→ A explicitly hands off to B
→ B replies in the same thread
→ B also has thread message state
→ human replies without a mention
→ both A and B may satisfy `start_only` ownership
```

That can wake both Agents and produce duplicate or conflicting replies.

For multi-Agent collaboration, `start_only` needs durable, exclusive active-owner state rather than historical “has this Connection ever posted in this thread?” state.

Recommended behavior:

```text
A is active owner
→ A explicitly hands off to B
→ B becomes active owner
→ unmentioned human follow-ups go only to B
→ B explicitly mentions A to hand work back
```

Agent A’s private session remains intact but inactive. It should not receive new messages until explicitly mentioned or made active owner again.

If ownership is absent or ambiguous, fail closed to explicit mentions rather than fan out.

## 5. Explicit-mention mode needs defined multi-Agent semantics too

`every_message` is safer because each delivery has an explicit target. It avoids accidental activation based on historical thread participation.

It still does not create shared runtime context:

```text
A mentions B
→ B receives only B's new inbound delivery and resumes B's session

B replies in the same Slack thread
→ A sees it in Slack, but does not receive it in A's runtime session

B mentions A again
→ A receives that explicit handback and resumes A's session
```

Please define the behavior for a message that explicitly mentions multiple Agents:

- Human multi-mentions may be supported as intentional parallel fanout, if that is a product feature.
- Agent-to-Agent handoffs should target exactly one recipient by default.
- Deliberate multi-Agent fanout and result synthesis should use a future coordinator/orchestrator model rather than accidental multiple mentions.

## 6. Context transfer must be explicit and bounded

Sharing a Slack thread must never imply sharing private Agent context.

A handoff should provide a bounded, auditable packet:

```text
From: Researcher
To: Writer
Task: Turn the findings into a customer-ready summary.
Context: Research found X, Y, and Z.
References: linked Slack messages, documents, or issues
```

The recipient may optionally receive a bounded excerpt of Slack-visible thread content, but must not automatically inherit:

- the previous Agent’s private runtime history;
- hidden instructions;
- tool outputs;
- long-lived memory;
- unrelated prior context.

This preserves clear Agent boundaries and avoids privacy, token-budget, and instruction-provenance problems.

## 7. Fix the legacy self-message path

The commit promises that an Agent’s own messages are always ignored. That appears true for modern Slack events, but the legacy fallback uses `event.bot_id` while ingress captures the bot user ID. Those Slack IDs are distinct.

A legacy self-authored event could bypass the self-message check and be accepted if it mentions the Agent’s own bot user ID.

Please capture and compare both identities, or fail closed for legacy bot events when self identity cannot be reliably determined.

## 8. Add end-to-end coverage

The plugin unit tests are useful, but this changes a Communications business rule and needs gateway/integration coverage for:

- persisted `accept_agent_mentions` enabled and disabled;
- peer Agent A → Agent B handoff using resolved peer identity;
- unmentioned bot message rejected as `mention_required`;
- external/unknown bot behavior matching the chosen trust policy;
- modern and legacy self-authored events rejected;
- `start_only` active-owner transfer and no duplicate wakeup;
- `every_message` handback behavior;
- multi-target explicit mention behavior;
- organization and Slack-workspace isolation;
- recipient runtime context containing safe handoff sender metadata.

## Acceptance criteria

- An Agent can intentionally mention an eligible peer without pre-existing knowledge of its Slack ID.
- Outbound handoffs render a real Slack provider mention.
- The recipient can identify the sender and handoff provenance.
- Peer resolution is tenant- and workspace-scoped.
- Unknown bots and self-authored messages fail closed.
- `start_only` has exactly one active Agent owner per thread.
- Explicit-mention mode has documented multi-recipient behavior.
- Agents retain separate private sessions; context moves only through an explicit, bounded handoff.
- Existing human mention and direct-message behavior remains unchanged.

Until peer resolution, context-transfer semantics, and thread-ownership rules are in place, this should be described as opt-in admission of explicitly addressed bot messages not complete Agent-to-Agent collaboration.

yeabtsega07:
Thanks for the review. Pushed a follow-up commit that addresses the parts I can close in this PR, and narrowed its claim to match:

- **Point 7 (legacy self-message gap): fixed.** Legacy bot events that carry only `bot_id` now fail closed as `bot_ignored`, as does any bot event where ingress did not capture this Agent's bot user id. The `bot_id` sender fallback is gone. Tests cover the toggle on and off.
- **Point 2 (trust boundary): documented as a security decision** rather than implemented as a peer allowlist. The setting is renamed to `accept_bot_mentions` / "Mentions from other bots", its description states that any Slack bot in a permitted channel can then address the Agent, it stays off by default, and the docs say it is meant for Agents in controlled channels.
- **Rename:** title, docs, and change log now describe this as opt-in admission of explicitly addressed bot messages, not Agent collaboration.

Points 1, 3, 4, 5, 6 (peer discovery, provenance to the recipient, exclusive `start_only` ownership, multi-recipient semantics, bounded context transfer) set multi-Agent semantics for the whole platform. I'd propose those as follow-up design work owned by your team; I'm happy to contribute what we learned. Our pipeline runs on `every_message` with explicit single mentions and does not depend on them. Point 8's gateway-level tests I can add here if you want them for the narrowed scope, say the word.

Checks: `make check-api`, `make check-migrations`, and the full `make test-api` suite were green before this commit; after it, check-api, the communications suites, and the full `make test-api` suite (1916 passed) are green as well.

yeabtsega07:
Update: after discussing with Kalkidan, our pipeline will hand work between agents through Jira (one agent files issues, the others poll on a cron), so this PR is no longer needed for our use case. Leaving it open for the maintainers to close or keep as a reference for the real agent collaboration design. Thanks for the thorough review.

PR #183, opened 2026-09-08
AF-291 fix(ui): refresh the org switcher after deleting an organization

### LINKED ISSUE #153: Deleted organization remains in the list until page reload
When a user deletes an organization, the organization is successfully deleted from the backend, but the UI list does not update automatically. The user has to manually refresh the page to see it disappear.

- Expected Behavior: The UI should immediately filter out or remove the deleted organization from the list upon a successful API response, without requiring a page reload.

ddlaurin:
Closes #153 

## Problem

The org switcher gets its list from the user's memberships in
current-user-context. `useDeleteOrganization` invalidates the two organizations
list queries but not that one, so a deleted org stays in the switcher until you
reload the page. Clicking it gives "You do not have access to this organization".

`useCreateOrganization` already invalidates current-user-context. Delete just missed it.

## Change

Invalidate current-user-context on delete, same as create.

## Testing

`organizations.spec.ts` already covered delete, but only asserted the toast and
the redirect so it passes with and without this bug.

The new test fakes the user endpoint so it returns two orgs before the delete
and one after, the way the real API would. Then it deletes an org and checks the
switcher drops it without a reload. It fails without the fix and passes with it.

## Checks

`pnpm exec tsc --noEmit` clean
`pnpm lint` clean
`pnpm exec playwright test` 252 passed

### DIFF
--- a/ui/src/features/organizations/components/organization-detail.tsx
+++ b/ui/src/features/organizations/components/organization-detail.tsx
@@ -23,6 +23,7 @@ import {
   DialogTitle,
 } from "@/components/ui/dialog";
 
+import { useOrgStore } from "../stores/org-store";
 import { useDeleteOrganization } from "../hooks/use-organization-actions";
 import { useOrganization } from "../hooks/use-organization";
 import { useOrganizationMembers } from "../hooks/use-organization-members";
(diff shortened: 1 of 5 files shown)


AnaniyaT:
This refetch of `/auth/me` runs concurrently with the `router.push` in `organization-detail.tsx`. If the refetch resolves first, `OrganizationProvider` sees the URL org is gone and redirects to the fallback org, overriding the push. Worth checking whether that is a problem here.

ddlaurin:
Reproduced. Tried moving the invalidation to `onSettled`, but it still ran before the route change. Also tried putting it to `organization-detail` but the page unmounts when the navigation completes, so the effect is gone before the route changes. 
Ended up using the org store to temporarily save the org id that's being deleted. `OrganizationProvider` skips the fallback redirect while the URL still points at that org, and clears the marker once the route has moved on.

AnaniyaT (approved):

PR #184, opened 2026-09-08
AF-290 fix(ui): collapse the top nav into a drawer on mobile and tablet screens

### LINKED ISSUE #154: Navbar is not mobile responsive and causes horizontal scrolling
The main navigation bar is not mobile-responsive. On mobile viewports (widths below 768px), the navbar elements do not collapse or scale down. This forces the entire page to scroll horizontally, breaking the user experience.

Steps to Reproduce
1. Open the application on a mobile device or use browser developer tools to simulate a mobile screen (e.g., iPhone/Pixel dimensions).
2. Look at the top navigation bar.
3. Try to swipe left and right on the page.

Expected Behavior
The navbar should adapt to smaller screens (e.g., collapse into a hamburger menu or stack vertically). The page layout should remain locked vertically, with **zero horizontal scrolling**.

Actual Behavior
The navbar elements maintain their desktop widths, spilling off the right side of the screen and causing the entire page to scroll horizontally.

ddlaurin:
Closes #154 

## Problem

The header is a single flex row whose children default to `min-width: auto`, so
nothing could shrink below its content width. Platform view needs ~1079px to lay
its seven tabs out inline, so a 375px viewport forced the document to 962px and
the whole page scrolled sideways.

## Change

- Tabs collapse into a `Sheet` drawer below `lg` (1024px), covering phones and tablets
- Header spacing tightens between `lg` and `xl` so all seven tabs fit at 1024px
- Drawer open state is derived from the current route, so back/forward, the org
  switcher and programmatic navigation all close it
- Brand wordmark and the "Hire agent" label collapse below `sm`

## Checks

- `pnpm exec tsc --noEmit` clean
- `pnpm lint` clean
- `pnpm exec playwright test` 251 passed (including 8 new specs)

### DIFF
--- a/ui/src/components/top-nav.tsx
+++ b/ui/src/components/top-nav.tsx
@@ -54,7 +54,10 @@ export function TopNav({ onHire }: TopNavProps) {
         { href: `${orgBase}/settings`, label: "Settings" },
       ];
   const [menuOpen, setMenuOpen] = useState(false);
-  const [navOpen, setNavOpen] = useState(false);
+  const routeKey = pathname ?? "";
+  const [navOpenedOn, setNavOpenedOn] = useState<string | null>(null);
+  const navOpen = navOpenedOn !== null && navOpenedOn === routeKey;
+  const setNavOpen = (open: boolean) => setNavOpenedOn(open ? routeKey : null);
   const menuRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


AnaniyaT:
`isActive` prefix-matches, so Overview counts as active on every platform page. It was already like that inline, but the drawer's background fill makes both rows look selected. An exact match for the overview route would fix it, like Home does in org view.

AnaniyaT (approved):

PR #190, opened 2026-09-10
AF-174: per-agent cost trend and an agents-by-spend table

ddlaurin:
## Problem

An agent's cost was a single total with no way to see how it trended, and there
was no per-agent breakdown of an organization's spend.

## Change

**Agent About tab** replaces the "Coming soon" placeholder with:

- spend over time, reusing `SpendOverTimeChart`
- the same `DateRangePicker` the costs page uses, with the range in the URL
- totals, and a per-model breakdown table, which the endpoint already returned

**Organization costs page** adds "Agents by spend": every agent with spend in
the window, sortable on any column. Selecting a row filters the page to that agent.

**Backend**

- `AgentCostRead` gains `spend_over_time` plus the resolved window, built from
  the existing series query under the Agent-pinned filter `get_agent_cost` already used.
- New `GET /organizations/{org}/costs/agents`, mirroring `spend_by_organization`.
  Uncapped, unlike the chart's series: a line per agent stops being readable
  after a handful, but a table has to account for every agent that spent anything.

## Design notes

- The table sorts client-side. The endpoint returns every agent rather than
  a page, so re-sorting reorders data already held rather than refetching.
- `GET /costs/agents` accepts a `sort` from the shared filter and ignores it;
  the ranking is always by spend. That matches `/summary` and `/filters/*` routes.

## Testing

- `make check-api`, `make check-migrations` pass
- `pytest api/tests/integration/test_costs.py` 24 pass, 4 new
- `pnpm exec tsc --noEmit`, `pnpm lint` clean
- `pnpm exec playwright test` 261 pass, 11 new

### DIFF
--- a/ui/src/features/agents/components/about-tab.tsx
+++ b/ui/src/features/agents/components/about-tab.tsx
@@ -1,7 +1,7 @@
 "use client";
 
-import { useState } from "react";
 
+import { DateRangePicker } from "@/components/date-range-picker";
 import { Skeleton } from "@/components/ui/skeleton";
 import { ApiError } from "@/shared/api/error/errors";
 import { SpendOverTimeChart } from "@/features/costs/components/spend-over-time-chart";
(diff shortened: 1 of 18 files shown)


ui/src/features/agents/components/about-tab.tsx
```diff
@@ -1,13 +1,171 @@
...
+            </p>
+          </div>
+
+          <DateRangePicker
+            from={dateFilters.from}
+            to={dateFilters.to}
+            onChange={(from, to) =>
+              setDateFilters({ from: from || null, to: to || null })
+            }
+            placeholder="All dates"
```
AnaniyaT: Empty picker sends no period, so the server uses its 30-day default, but the label says "All dates". "Total spend" then reads as lifetime and isn't. The response already returns from_date/to_date/granularity; render those instead of the static placeholder, and drop "by day" on line 38 since granularity varies with the span.

ui/src/features/costs/format.ts
```diff
@@ -72,3 +72,10 @@ export function formatDuration(ms: number | null): string {
   if (ms < 1000) return `${ms}ms`;
   return `${(ms / 1000).toFixed(1)}s`;
 }
+
+/** The model as the cost surfaces show it: the slug alone, without its routing path.
+ *  `cost_record.model` stores the full route (e.g. "openrouter/anthropic/claude-opus-5"),
+ *  which is too long to read in a table and identical across rows in its leading parts. */
+export function formatModelLabel(model: string): string {
```
AnaniyaT, line 79: Duplicate of `shortModel` in cost-summary-cards.tsx, and cost-row.tsx inlines the same expression. Keep this one and point the other two at it.

AnaniyaT (approved):

PR #191, opened 2026-09-10
feat(costs): add opt-in organization LLM budgets

tadas-subonis:
Organizations currently have per-Agent LiteLLM keys but no shared allowance. This adds one automatically provisioned LiteLLM team per Organization and an optional deployment-configured USD budget shared by its Agent keys. Enforcement remains disabled when `ORGANIZATION_LLM_BUDGET_USD` is unset or empty; `ORGANIZATION_LLM_BUDGET_DURATION` defaults to `30d`.

Both self-service Organization creation and platform user provisioning call the LiteLLM client directly to create teams. The existing OrganizationService handles deployment budget synchronization; no separate LLM domain service is introduced. New Agent keys require successful team provisioning and include `team_id`. API startup reconciles Organization team budget settings: amount-only updates preserve the renewal date, duration changes reschedule it, and removing the amount clears existing limits without resetting spend.

Existing Agent keys are left untouched. Their initial team assignment will be handled separately with a one-off operational script; this PR contains no key backfill or backfill tests. Legacy keys must be enrolled before team limits apply to them.

Helmfile and both deployment workflows accept separate k3s, staging, and public budget variables. No deployment amounts are invented or enabled by default. The operations guide documents configuration, legacy enrollment, disabling limits, and the distinction between LiteLLM budget counters and healed reporting costs. The API chart packaging version is 0.9.1. No schema migration or new dependency is needed.

Validation:
- `make check-api`, `make check-migrations`, and Helm lint pass after merging the latest staging (2026-09-15).
- 61 relevant unit/integration tests pass after the staging merge, covering both creation paths, policy updates/removal, renewal preservation, zero/invalid settings, concurrency, and startup failure.
- Helm lint and `git diff --check` pass after the reduction.
- Before removing backfill, the full API suite passed (2,068 tests), along with Helmfile environment mapping, Helm renders for unset/zero/positive budgets, and the GitHub Kubernetes integration check.

Rollout requires setting `ORGANIZATION_LLM_BUDGET_USD` for k3s main and `PUBLIC_ORGANIZATION_LLM_BUDGET_USD` for the public deployment, then deploying. Budgets use LiteLLM-recorded spend, not an exact provider-invoice ceiling. Historical pre-enrollment requests are not charged to the new team counter.


aai-labs-agent (commented):
## Code Review — feat(costs): add opt-in organization LLM budgets

**Verdict: Comment** (1 suggestion, no blocking issues)

### Summary

This PR adds an optional deployment-configured USD budget per Organization via automatically provisioned LiteLLM teams. The design is clean: teams are provisioned on org creation (both self-service and platform-user paths), reconciled at API startup, and new agent keys require a team before issuance. Existing keys are deliberately left untouched (deferred to an operational script), which is called out clearly in the PR body and docs.

The reconciliation logic in `ensure_organization_team` is well thought out — it patches only changed fields, preserves spend and renewal dates, handles concurrent team creation via re-read verification, and fails closed on incomplete provisioning. The `field_validator` that treats empty strings as `None` (unlimited) while `0` remains a zero allowance is a nice distinction.

### Findings

**Suggestion — `api/domains/organizations/service.py:66`** — `sync_llm_budgets` propagates without identifying the failing org. Inline comment with a suggested fix.

### Looks Good

- Reconciliation is idempotent and safe to retry — an unchanged policy performs no update, so restarts don't reset spend or renewal windows.
- `ensure_organization_team` handles the concurrent-creation race correctly: a 400/409 from `/team/new` is followed by a re-read rather than assuming success.
- `generate_key` now calls `ensure_organization_team` before key generation and includes `team_id` — an agent can't get an unassigned key.
- Both creation paths (self-service `POST /organizations` and platform-user provisioning) handle LiteLLM failure gracefully: the committed org is not rolled back, and provisioning is retried on first key generation.
- Config validation is solid: `ge=0`, `allow_inf_nan=False`, and a regex pattern for duration prevent invalid values at startup.
- The `field_validator` that maps empty/whitespace strings to `None` (unlimited) while preserving `0` as a zero allowance is the right semantic distinction.
- Helm chart version bumped (0.9.0 → 0.9.1) and `appVersion` is absent — consistent with the repo convention for charts that build custom images (no `appVersion` field present).
- Documentation is thorough: `docs/features/costs.md`, `docs/guidelines/operations.md`, `docs/features/identity-and-organizations.md`, `docs/architecture/api.md`, and `docs/INDEX.md` all updated in the same diff, satisfying the AGENTS.md review protocol requirement.
- The operations guide clearly documents the three deployment environments (k3s, staging, public) with their separate GitHub Actions variables, legacy key enrollment, and the distinction between LiteLLM budget counters and healed reporting costs.
- Test coverage is comprehensive: 194 unit tests covering config validation, team creation, budget patching, concurrent creation, failed key generation, and startup reconciliation; 77 integration tests covering both creation paths and retry behavior.

### Suppressed

- **No `appVersion` in Chart.yaml**: Checked — the chart has no `appVersion` field at all (only `apiVersion`, `name`, `description`, `version`). The version bump from 0.9.0 to 0.9.1 is correct since `deployment.yaml` templates changed. No `appVersion` gap to flag.
- **`list_ids` opens its own Session**: Matches the established pattern in `OrganizationRepository` where read methods like `get_read` and `get_platform_read` each open their own `Session`. Not a convention violation.
- **Exception swallowing in creation paths**: Both `create_organization_for_current_user` and `create_platform_user` catch all exceptions from `ensure_organization_team` and log them without re-raising. This is intentional per the PR body — the transaction is already committed and key generation retries provisioning. The design is sound.

---
*Reviewed by Hermes Agent*

api/domains/organizations/service.py
```diff
@@ -50,12 +51,21 @@ def _and_list(items: list[str]) -> str:
...
     event_delivery_dispatcher: EventDeliveryDispatcher
     agent_settings_lookup: AgentSettingsLookupService
     agent_repository: AgentRepository
 
+    def sync_llm_budgets(self) -> None:
+        config = get_config()
+        if not config.litellm_base_url or not config.litellm_secret_name:
+            return
+        for organization_id in self.organization_repository.list_ids():
+            self.litellm.ensure_organization_team(str(organization_id))
```
aai-labs-agent, line 66: If one org's reconciliation fails, the exception propagates without identifying which org caused it. An operator debugging a startup crash sees only `Organization LiteLLM reconciliation failed (LiteLLMError)` with no org_id to investigate.

Wrapping the loop body to log the org_id before re-raising would make this debuggable:

```python
for organization_id in self.organization_repository.list_ids():
    try:
        self.litellm.ensure_organization_team(str(organization_id))
    except Exception:
        logger.error("LiteLLM reconciliation failed for Organization %s", organization_id)
        raise
```

aai-labs-agent:
**Correction:** In the review above I wrote "271 new tests" — that's incorrect. 271 is the number of added test lines across the two new test files. The actual test count is 13 test functions expanding to 29 parametrized cases (24 unit + 5 integration) via `pytest.mark.parametrize`.

PR #195, opened 2026-09-10
Af 292 capture and restore agents volume

dominykas-aai-labs:


api/domains/agents/builders/restore_point.py
```diff
@@ -0,0 +1,192 @@
...
+SOURCE_MOUNT_PATH = "/source"
+DEST_MOUNT_PATH = "/dest"
+TARGET_MOUNT_PATH = "/target"
+BACKUP_MOUNT_PATH = "/backup"
+ARCHIVE_MOUNT_PATH = "/archive"
+
+_ENTRYPOINT = ["python", "-c", "from api.domains.agents.restore_point_job import main; main()"]
+
+_TTL_SECONDS_AFTER_FINISHED = 3600
+_BACKOFF_LIMIT = 1
```
AnaniyaT: `backoff_limit=1` applies to the restore job too, and `restore()` re-runs its pre-restore capture on attempt 2 — against the already-wiped target, truncating the good backup in `"w:gz"`. Either `backoff_limit=0` for restore, or skip the capture when `backup/manifest.json` exists. Worth a test on the second attempt; `test_capture_job_does_not_restart_or_retry_indefinitely` only covers capture.

api/domains/restore_points/service.py
```diff
@@ -0,0 +1,531 @@
...
+                target.id,
+                RestorePointStatus.RESTORING,
+                from_statuses=(RestorePointStatus.READY,),
+                event_name=AGENT_RESTORE_POINT_RESTORED,
+                actor=resolve_actor_identity(context, current.organization_id),
+                payload=self._event_payload(current, target, context),
+            )
+            if result is None:
+                raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=NOT_READY_DETAIL)
+            self.repository.set_job_name(target.id, job_name)
```
AnaniyaT: `job_name` is set in a second transaction after the `RESTORING` transition, and `mark_ready` nulled it — so a concurrent reconcile sees `job_name is None`, takes `_fail_capture` (`service.py:119`), and deletes the archive PVC being restored from. It compounds from there: `set_job_name` then matches nothing because the row is terminal, its `False` return is ignored, and `_provision_restore` creates a Job mounting a PVC that no longer exists — so the pod sits unschedulable until `activeDeadlineSeconds`, blocking agent start and delete for 30 minutes. Folding `job_name` into the `update_status_with_event` values closes it.

api/domains/restore_points/service.py
```diff
@@ -0,0 +1,531 @@
...
+
+        for row in rows:
+            try:
+                self._reconcile_row(row)
+            except Exception:
+                logger.warning("Could not reconcile restore point %s", row.id, exc_info=True)
+
+    def _reconcile_row(self, row: AgentRestorePoint) -> None:
+        namespace = self.config.k8s_namespace
+        if not row.job_name:
```
AnaniyaT, line 157: This branch always calls `_fail_capture`, which deletes the volume — wrong for a `RESTORING` row, whose archive is intact and is what a retry reads from, and contrary to `_fail_capture`'s own docstring. Only reachable via the race above today, but worth branching on status the way the `job is None` case below already does, plus a test asserting that row keeps its PVC.

api/domains/agents/restore_point_job.py
```diff
@@ -0,0 +1,224 @@
...
+
+class DestinationFullError(Exception):
+    pass
+
+
+def _matches_prefix(rel_path: str, prefixes: tuple[str, ...]) -> bool:
+    return any(rel_path == prefix or rel_path.startswith(prefix + "/") for prefix in prefixes)
+
+
+def _is_regenerated_openclaw_markdown(rel_path: str) -> bool:
```
AnaniyaT: This drops every agent-authored `workspace/*.md` on OpenClaw while Hermes keeps it — `test_hermes_capture_retains_agent_owned_memories_and_work` and `test_openclaw_capture_retains_workspace_user_md_but_drops_other_markdown` assert both halves. `init-openclaw.js` copies a known eight-file list; mirroring it would remove the divergence.

api/tests/integration/test_kubernetes_client.py
```diff
@@ -118,6 +123,23 @@ def _config_map(name, run_id):
     )
 
 
+def _job(name, run_id):
```
AnaniyaT: Worth adding three cases here — this is the only file `make test-api-k8s` runs against real k3d, and the new log/exit-code readers have no cluster coverage at all. The existing busybox fixture covers most of the setup:

- `get_job_exit_code` with two pods (`backoff_limit=1`, `sh -c 'exit 3'`) — assert the newest pod wins.
- `get_job_exit_code` on a job killed by `activeDeadlineSeconds` — assert it returns `None`, so the service has a documented value to branch on.
- `read_job_logs` / `get_pod_name_for_job` on a job that prints a JSON line then exits non-zero — covers newest-pod selection and manifest parsing.

Service-side interpretation stays in `test_agent_restore_points.py` on `MockK8sModule` as it is now; the k3d cases just pin down what those mocks should return.

AnaniyaT (changes requested):
Requesting changes on three things: a retry path that can silently destroy the safety net, a race that can permanently delete the archive being restored from, and a runtime asymmetry that needs a decision rather than a fix. Restating the changelog's own condition — Ticket 2 needs to land in the same release or this ships a storage leak.

AnaniyaT (approved):

PR #199, opened 2026-09-12
AF-269: Surface Agent provisioning failures in the UI

ddlaurin:
## Problem

When Kubernetes rejected an Agent's resources during provisioning, the Agent was persisted in `ERROR` with a useful `last_error`, but nothing reached the user. The start endpoint answered `500 {"detail": "Failed to start agent <uuid>"}` and the Agent read carried no failure at all, so the cause was only reachable through database or cluster diagnostics.

Observed in staging while creating a Discord Agent whose PVC was refused by an exhausted namespace storage quota.

## What changed

`api/domains/agents/provisioning_errors.py` normalizes any exception raised while provisioning into one of five categories — quota exhaustion, cluster RBAC denial, rejected manifest, unreachable cluster, or an unclassified failure. Each category states what the reader can do: a failure only an administrator can clear says so, one worth retrying says to try again.

No raw error text reaches a client. The only variable part is `detail`, rebuilt from fragments matching strict patterns (e.g. a Kubernetes quantity, a resource kind); anything else is dropped. A denylist gate behind that catches a fragment that would otherwise carry a credential. The full exception still goes to the API logs.

**The contract**, returned identically as a failed start's error body and as `last_error` on the Agent read while it is in `ERROR`:

```json
{
  "code": "QUOTA_EXHAUSTED",
  "category": "quota_exhausted",
  "summary": "The agent could not start — its namespace has run out of resource quota. Ask an administrator to free up or raise it, then start the agent again.",
  "detail": "requests.storage: requested 1Gi, used 30Gi, limit 30Gi (quota example-quota, creating persistentvolumeclaims)"
}
```

HTTP status follows the category: 503 for quota, cluster RBAC and unreachable cluster; 500 for a rejected manifest or an unclassified failure.

Previously only Kubernetes resource creation was wrapped, so a failure earlier in provisioning appeared as an unexplained 500 and left the Agent looking untouched. The whole of `_provision_and_start` is now covered. Deterministic preconditions opt out explicitly by raising `AgentProvisioningPrecondition` (already running, no pinned template, a model the Organization no longer allows, and an incompletable Google Workspace credential).

`summary` is derived from the stored `code` at read time, so improved copy reaches Agents already in `ERROR` without a data migration. A row with no code predates this contract; its text came from the cluster, so it reports as unclassified and the stored text is dropped.

The Agent detail page shows the failure in a banner read from the Agent itself, so it paints on first render and survives a refresh until a successful start clears it. Hire, start, and Apply & Restart all report the same classified failure. `handleError` now reads a message and code out of a `detail` object.

Creation can succeed while the start fails. The dialog now holds the created Agent, hides the creation form, and offers `Start again`, which starts that Agent. Previously a retry created a second Agent and a second LiteLLM key.

## Deliberately not implemented

Two classifications were dropped after testing against a real cluster:

- **Image unavailable** — Kubernetes accepts a Deployment naming an unpullable image and fails it later on the pod. It never reaches the provisioning path. The existing health surface already reports it as `ErrImagePull`.
- **Resource conflict** — every creation goes through `_create_or_get`, which resolves a 409 by reading the existing object, so a conflict never propagates.

A 5xx from the API server is classified as unclassified rather than "namespace unreachable": the server answered, so the namespace is reachable.

## Migration

`c8e3a5b71d94` adds two nullable columns, `agent.last_error_code` and `agent.last_error_detail`.

## Testing

- 17 unit tests covering classification, the quota axis extraction, RBAC detail without the service account identity, and redaction of Secret values, bearer tokens and kubeconfig content
- 7 integration tests covering the end-to-end contract, persistence across reads, clearing on a successful start, failures raised before any cluster call, and cross-organization concealment
- 8 Playwright tests for the banner, refresh persistence, immediate mutation failure, Apply & Restart, and fallback copy; plus hire-dialog coverage for the provisioning alert, the retry, and the hidden form

Verified manually against a local k3d cluster by exhausting a namespace storage quota, by denying the service account RBAC, and by taking the cluster offline.

## Out of scope

`create_agent`, `stop`, and `get_agent_health` each reach Kubernetes outside this wrapper — `create_agent` reads the LiteLLM master key from a Secret, so an unreachable cluster still returns an unclassified 500 from those three paths.


ui/src/features/agents/components/hire-dialog.tsx
```diff
@@ -228,6 +273,35 @@ export function HireDialog({ onClose, onHired }: HireDialogProps) {
             />
           </section>
         )}
+          </>
+        )}
+
+        {provisioningFailure && (
```
dominykas-aai-labs, line 288: some start failures show no message at all

the template has required skills and the user entered credentials. Create succeeds, then start fails with something that isn't a provisioning failure: the 400 Google Workspace client id/secret precondition, the 409 lifecycle lock, or a network error. The user sees only "Start again", with no explanation.
The "was created and is waiting on your team page" note also only appears for provisioning failures.

api/domains/agents/provisioning_errors.py
```diff
@@ -0,0 +1,328 @@
...
+    # Reporting quota exhaustion as an RBAC fault sends operators to the service
+    # account when the namespace has simply run out of a resource, so quota wins.
+    if "exceeded quota" in lowered or "resourcequota" in lowered:
+        return AgentProvisioningErrorCategory.QUOTA_EXHAUSTED
+    if status_code == 403 or "forbidden" in lowered or "cannot create resource" in lowered:
+        return AgentProvisioningErrorCategory.CLUSTER_PERMISSION_DENIED
+    if status_code == 422 or "is invalid" in lowered or "unprocessable" in lowered:
+        return AgentProvisioningErrorCategory.RESOURCE_REJECTED
+    if status_code == 404 or "not found" in lowered:
+        return AgentProvisioningErrorCategory.CLUSTER_UNAVAILABLE
```
dominykas-aai-labs: Substring matching now runs on str(exc) for any exception, not just Kubernetes ones, because the try/except covers the whole provisioning step. For example, RuntimeError("skill version not found") becomes "cluster unreachable, try again". The "forbidden" and "is invalid" checks have the same problem.

"resourcequota" in lowered also sends an RBAC denial on the resourcequotas resource to QUOTA_EXHAUSTED.

api/domains/agents/service.py
```diff
@@ -1864,21 +1903,50 @@ def start_agent(self, agent_id: UUID, context: CurrentUserContext) -> AgentRead:
...
+            raise
+        except Exception as exc:
+            logger.exception("Failed to start agent %s", agent.id)
+            raise self._record_provisioning_failure(agent, exc) from exc
+
+    def _record_provisioning_failure(self, agent: Agent, exc: Exception) -> HTTPException:
+        """Persist a sanitized failure on the Agent and build the error to raise.
+
+        The full exception is already in the logs; only the normalized form is
+        persisted or returned, so no cluster text reaches a client.
```
dominykas-aai-labs, line 1937: The try/except now covers save_with_lifecycle_event too. If that DB write fails after every Kubernetes resource was created, the Agent is recorded as ERROR (PROVISIONING_FAILED) while its pod is running.

The repository.save in the handler can also raise on the same DB fault. That hides the original exception and returns a bare 500 without the classified body.

ui/src/shared/api/error/error-handler.ts
```diff
@@ -2,17 +2,42 @@ import axios from "axios";
...
 import { ApiError } from "./errors";
 
+// FastAPI sends `detail` as a string, a list of validation errors, or an object
+// such as a classified provisioning failure or an agent health report.
+const isDetailObject = (detail: unknown): detail is Record<string, any> =>
+  typeof detail === "object" && detail !== null && !Array.isArray(detail);
+
+const messageFrom = (data: any, fallback: string): string => {
+  const detail = data?.detail;
+
```
dominykas-aai-labs, line 12: messageFrom in ui/src/shared/api/error/error-handler.ts already sets ApiError.message = detail.summary, so toastError(cause) already shows the summary. That makes the following redundant:

- reportLifecycleFailure in agent-detail-page.tsx: revert to .catch(toastError)

- provisioningFailureLine in provisioning-failure.ts: it only returns failure.summary, so delete it

- the restart() rewrap in use-agent-apply-and-restart.ts: it also replaces the ApiError with a plain Error, losing status and code, so revert it

- the onError invalidation in use-create-agent.ts: create never provisions (only start_agent and maintenance call _start_agent_unchecked), so a failed create changes nothing

- the hire dialog's inline error panel markup, which duplicates AgentErrorBanner: have the banner accept children (for the "was created" line) and a className, then reuse it

docs/features/agents.md
```diff
@@ -22,6 +22,8 @@ An Agent is the central execution aggregate. It connects organization tenancy, a
...
+- Agent read DTOs expose the normalized failure as `last_error` while the Agent is in `ERROR`, and a failed create/start/restart returns the same shape as its error body, so a client never has to derive the cause from a status code. Both are rebuilt from the stored category at read time rather than trusted from storage, so a row written before this contract existed reports an unclassified failure instead of its unsanitized text. `last_error` is subject to the same Agent Access rules as the rest of the Agent read.
```
dominykas-aai-labs: "a failed create/start/restart returns the same shape". Create never provisions. The hire flow is create followed by start. Reword to "start/restart (including the start that follows hiring)".

dominykas-aai-labs:
ui/src/features/agents/components/agent-configuration-page.tsx:143-158

The ERROR alert still reads health?.reason, which needs activity.read, and shows "Runtime reported: The agent could not start — …". The text now duplicates itself, and users without activity.read get generic text even though agent.lastError is on the Agent.
Fix: render AgentErrorBanner (or its title, summary and detail) from agent.lastError when present, and fall back to the current copy otherwise.

dominykas-aai-labs (changes requested):

dominykas-aai-labs:
### Should fix before merge

**1. Restore point failures show start-failure wording**
`api/domains/restore_points/service.py` (lines 453 and 577) saves `friendly_k8s_error(exc)` as the failure reason. That function now returns the start-failure copy, so a failed backup or restore reads "The agent could not start — …". Most non-Kubernetes errors get the generic "creating its runtime resources failed unexpectedly" text. This only showed up after staging was merged in. Restore points need their own wording, or a classifier call that doesn't use the start-specific copy.

**2. A missing namespace is still described as a temporary outage**
`api/domains/agents/provisioning_errors.py` (`_classify`) still maps a Kubernetes 404 to `CLUSTER_UNAVAILABLE` ("often temporary, so try again"). A missing namespace needs an administrator, not a retry. It should get its own copy that points to an administrator. The non-Kubernetes and `resourcequotas` parts of the earlier comment are fixed.

**3. No test for a same-org user without Agent Access**
`test_a_failure_on_an_inaccessible_agent_is_not_readable_across_organizations` only covers an agent in another organization. The acceptance criterion is about Agent Access, so please add a case for a member of the same organization with no access assignment to the agent.

**4. No test for a failed Restart**
The new Restart menu has no e2e test for a restart whose start fails. It should check that the toast appears and the banner shows after the refetch.

### Minor

- **Restart label:** between the stop and the start, the refetched agent is STOPPED, so the control switches to a disabled "Start" button instead of showing "Restarting…".
- **Quota detail allowlist:** it lacks the canonical `count/deployments.apps` and `count/jobs.batch` axes, so those quota failures show only the summary. This fails safe.
- **Configuration page:** when the new banner is shown, the "View Agent logs" link is gone.
- **Unverified:** a LimitRange rejection also comes back as 403 "forbidden", so it is likely reported as "Cluster permission denied". Worth checking on k3d with a PVC LimitRange (max storage 500Mi).

PR #205, opened 2026-09-16
AF-303: set LLM spend limits per Organization

AnaniyaT:
Gives each Organization its own LLM spend limit, set by a Platform Administrator, and the surfaces that make it visible before an Organization's agents stop working.

Builds on #191 and supersedes it — that branch is the base here, so its commits are included and it should be closed rather than merged. The team provisioning it added is kept; the deployment-wide budget is replaced by a per-Organization one, and the legacy key enrollment it removed is restored.

## What's included

- **Per-Organization limits** — `llm_budget_usd`/`llm_budget_duration` on `organization`, set through `PUT /platform/organizations/{id}/llm-budget`. An Organization cannot see or change its own: a cap a customer can raise is not a cost control. The row is authoritative and the proxy is a projection of it, so a save that cannot reach the proxy returns 502 with the setting kept.
- **Removes the deployment-wide budget** — `ORGANIZATION_LLM_BUDGET_USD`/`_DURATION` are gone from `.env.spec`, both deploy workflows, `values.yaml`, `deployment.yaml` and `helmfile.yaml.gotmpl`. the only budget lines left in `deployment.yaml` are the two for the alert thresholds.
- **Splits the remote team calls** — `ensure_team_exists` provisions identity only and is what key generation calls; issuing a key must never re-assert a spend policy its caller was not given. `apply_team_budget` writes policy, and both verify by re-reading rather than trusting the response.
- **Legacy key enrollment** — `POST …/llm-budget/enroll` and `GET …/llm-budget/coverage` return a per-agent census; one agent failing never ends the sweep, and a key already in another team is reported rather than moved. The budget controls stay hidden until every agent is covered, so a limit is never set over agents it would silently miss.
- **Reconciliation moved to a CronJob** — it previously ran in the API lifespan before serving traffic, so a hung proxy could hold the event loop until liveness killed the pod. The API no longer contacts the proxy at all.
- **Threshold alerts** — a second CronJob refreshes each capped Organization's spend snapshot every 5 minutes and emits domain events on a crossing. Alert state keys on the window *and* the limit, so raising a limit re-arms the thresholds rather than silencing the Organization for the rest of the period. Owners and Admins are notified; Platform Administrators only when an Organization is cut off.
- **Organization-facing visibility** — usage and limit on the Costs page behind `cost.read`, a warning banner there and an app-wide banner once exhausted. Members see nothing, so the agent-facing message is their only signal.
- **Agent-facing rejection** — both runtimes' in-pod LLM proxy now rewrites a budget rejection into something the person talking to the agent can act on, instead of the raw error naming an internal team id. Matched on the error body across 400 and 429 because the same statuses carry malformed requests, unknown models and rate limits, which keep their own errors.

Existing behaviour is unchanged where no limit is set, which is every Organization today: teams are still provisioned, keys still assigned, and nothing is enforced.

## Required repo config

`ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS` (GitHub Variable, optional) — comma-separated percentages, each 1–100, defaulting to `80,100`. A malformed list refuses to boot rather than quietly alerting nobody. The three budget variables #191 asked for are no longer needed and were never created.

## Notes

- Chart packaging version is `0.9.0`: a new template is more than a patch bump, and two CronJobs were added.
- Three migrations, all additive and reversible, plus a heads merge after taking staging.
- Left deliberately for separate changes, both pre-existing: `LiteLLMClient._master_key()` re-reads the Kubernetes Secret on every call with no caching, and `/key/info` carries the plaintext agent key in the query string. Both touch the shared client used by cost sync and key generation.
- The Costs page's totals and the spend allowance cover different windows by design, so both now name their period rather than appearing to contradict each other.
- `test_email_connection_is_refused_when_the_environment_has_no_agent_email_domain` fails locally for anyone whose `.env` sets `AGENT_EMAIL_DOMAIN`; it passes with the variable cleared and is unrelated to this change.

## Testing

`make check-api`, `make check-migrations`, `make check-ui`, `make lint-ui`, `make check-monitoring`, `helm lint helm/agentbarn-api` — all exit 0. `make test-ui` 296 passed. `make test-api` 2381 passed (1 pre-existing environment failure above). Helm renders verified with the alerting CronJob enabled and disabled.

Verified end to end against a real LiteLLM v1.96.2: a $0.01 limit blocked a live model call; three legacy keys enrolled and re-enrolling was a no-op; the alert event went through the outbox and worker to a `SUCCEEDED` delivery; both in-pod proxies were started for real (Python and node) against a stub upstream.

## Review round

All comments from the first review are addressed. The ones that changed behaviour:

- `ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS` is now exported by both deploy workflows — it was documented as a GitHub Variable but reached nothing, so every deployment silently got the default.
- `100` is always present in the alert thresholds. Only the highest crossed threshold fires, so a list like `50,90` left an exhausted Organization with a banner saying so and no notification.
- The enrollment gate fails closed: a failed coverage check used to leave the limit controls visible, which is the opposite of what the gate is for.
- A sweep no longer re-reads the Kubernetes Secret per Agent, nor re-reads a key's team the caller has just read; the coverage query has its own key and a `staleTime`, so a budget save no longer re-runs it.
- Agent keys are looked up by their SHA-256 rather than in plaintext — confirmed against a live v1.96.2 proxy, both forms resolve the same key.
- Budget, enrollment, coverage, alerting and reconciliation moved to `OrganizationLlmBudgetService`; `OrganizationService` is back to 381 lines.
- Everything a person can read says **spend limit**; `CONTEXT.md` carries the term. Identifiers and schema keep `llm_budget_*`.

Two suggestions I'd push back on rather than silently skip:

- **Splitting the in-pod error rewrite and the banners into their own PR.** A plain member sees no banner and gets no email, so the rewritten agent-facing error is their only signal that anything is wrong. Shipping limits without it reintroduces the silent failure this change exists to remove.
- **The unparenthesized `except (A, B)` groups.** `ruff format` strips those parentheses under Python 3.14, so that form is this repo's canonical style rather than an oversight.


api/migrations/versions/97f14c9b7000_merge_heads.py
```diff
@@ -0,0 +1,23 @@
...
+Revises: c3e6a91b47d2, c4e8a2f19d73
+Create Date: 2026-09-16 00:00:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+# revision identifiers, used by Alembic.
+revision: str = "97f14c9b7000"
+down_revision: str | Sequence[str] | None = ("c3e6a91b47d2", "c4e8a2f19d73")
```
kalibetre, line 13: **Required — two Alembic heads once this lands.** Staging has moved on (85 commits, 6 new migrations, including `d4f1a8c2e3b7`). Computing heads over the union of this branch and `origin/staging` gives `['d4f1a8c2e3b7', '97f14c9b7000']`, so `make check-migrations` fails and `upgrade head` is ambiguous. Merge staging again and add a merge revision.

helmfile.yaml.gotmpl
```diff
@@ -140,6 +140,8 @@ releases:
         value: {{ env "API_IMAGE_TAG" | default "" | quote }}
       - name: buildSha
         value: {{ env "GIT_SHA" | default "" | quote }}
+      - name: organizationLlmBudgets.alertThresholds
+        value: {{ env "ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS" | default "80,100" | quote }}
```
kalibetre, line 144: **Required — this variable never reaches Helm from CI.** `deploy.yml` and `deploy-public.yml` don't export `ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS` (compare `AGENT_MODEL_ALLOWLIST` at `deploy.yml:283` and `deploy-public.yml:326`; no workflow file is in this diff). The PR body advertises it as an optional GitHub Variable, but setting it is a no-op and every deployment gets `80,100`. Add `ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS: ${{ vars.ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS }}` to both workflows.

api/domains/organizations/service.py
```diff
@@ -50,12 +75,361 @@ def _and_list(items: list[str]) -> str:
...
+        organization.llm_budget_renews_at = _parse_timestamp(renews_at)
+
+        # The key spans the window and the limit, so both a renewal and a limit change
+        # re-arm the thresholds rather than leaving the Organization permanently quiet.
+        alert_key = f"{renews_at}|{limit}"
+        if organization.llm_alert_key != alert_key:
+            organization.llm_alert_key = alert_key
+            organization.llm_alerted_threshold = None
+
+        reached = [t for t in self._alert_thresholds() if limit > 0 and spend >= limit * t / 100]
```
kalibetre: **Should fix — a threshold list without `100` never fires the exhausted alert.** `highest = max(reached)`, so with `50,90` and spend at 100% `highest == 90`, which is already recorded, and `ORGANIZATION_LLM_BUDGET_EXHAUSTED` never fires. The "allowance reached" email (Owners, Admins and Platform Administrators) is never sent, while the banner still says exhausted (`get_organization_llm_budget`, `spend >= limit`), so the two surfaces disagree. `operations.md` invites any value from 1 to 100. It's latent only because of the workflow gap above. Force `100` into the list in the validator or `_alert_thresholds()`, and add a test — no service-level test uses a list without 100.

api/domains/organizations/service.py
```diff
@@ -50,12 +75,361 @@ def _and_list(items: list[str]) -> str:
...
+        self, encrypted_key: str, team_id: str, encryption_key: str, *, enroll: bool
+    ) -> tuple[AgentLlmEnrollment, bool]:
+        """Never lets one Agent's failure end the sweep, and never logs the key."""
+        try:
+            key = decrypt_token(encrypted_key, encryption_key)
+        except Exception:
+            logger.warning("Could not decrypt the stored LiteLLM key for an Agent in team %s", team_id)
+            return AgentLlmEnrollment.UNREADABLE, False
+        try:
+            current = self.litellm.get_key_team(key)
```
kalibetre: **Should fix — the coverage sweep does avoidable proxy work.** `_master_key()` reads the K8s Secret on every call (no caching). A coverage read costs one secret read plus one HTTP call per Agent; enroll costs four of each per unenrolled Agent, because this `get_key_team` is repeated by `attach_key_to_team` (`client.py`, once before the update and once after). Resolve the master key once per sweep and skip the redundant first read (e.g. let `attach_key_to_team` accept the already-known current team).

ui/src/features/organizations/hooks/use-organization-actions.ts
```diff
@@ -96,3 +99,78 @@ export function useUpdateOrganization({ toastOnError = true }: { toastOnError?:
...
+        `/api/v1/platform/organizations/${organizationId}/llm-budget`,
+        budget,
+        { schema: PlatformOrganizationSchema },
+      );
+      return response.data;
+    },
+    onSettled: () => {
+      // Refetched on failure too: a 502 here means the amount was stored and only the
+      // proxy push failed, so the card must not keep rendering the previous value.
+      void queryClient.invalidateQueries({
```
kalibetre, line 125: **Should fix — every budget save reruns the coverage sweep.** The coverage key is built from `platformOrganizationsKey.detail(organizationId)`, so invalidating that prefix here refetches `llm-coverage`, which fans out one proxy call per Agent. `useOrganizationLlmCoverage` also has no `staleTime`, so it refetches on every mount. Invalidate the detail key with `exact: true` (or move the coverage key out of that prefix) and give the coverage query a `staleTime`.

ui/src/features/organizations/components/llm-budget-card.tsx
```diff
@@ -0,0 +1,250 @@
...
+              inputMode="decimal"
+              placeholder="No limit"
+              value={amount}
+              disabled={busy}
+              aria-label="Spend limit in USD"
+              onChange={(event) => setAmount(event.target.value)}
+            />
+          </div>
+
+          <select
```
kalibetre: **Should fix — native `<select>`.** `docs/guidelines/webapp.md:7` says to use the shadcn `Select` primitives and not to add native `<select>` controls, with `SelectItem` inside a `SelectGroup`. Other native selects already exist in the repo, but the rule is not to add more.

ui/src/features/organizations/components/llm-budget-card.tsx
```diff
@@ -0,0 +1,250 @@
...
+  return `$${organization.llmBudgetUsd.toLocaleString()} ${durationLabel(organization.llmBudgetDuration)}`;
+}
+
+export function LlmBudgetCard({ organization }: { organization: PlatformOrganization }) {
+  const [amount, setAmount] = useState(
+    organization.llmBudgetUsd == null ? "" : String(organization.llmBudgetUsd),
+  );
+  const [duration, setDuration] = useState(organization.llmBudgetDuration ?? DEFAULT_DURATION);
+  const setBudget = useSetOrganizationLlmBudget(organization.id);
+  const { coverage, isLoading: coverageLoading } = useOrganizationLlmCoverage(organization.id);
```
kalibetre: **Consider — the enrollment gate fails open.** `gated` needs `coverage` to have loaded, and the hook's `error` is ignored here. If the coverage fetch fails, `coverage` is null, `gated` is false and the limit controls appear despite the documented "hidden until every Agent is covered" rule. Read `error` and keep the controls hidden (or show an inline retry) when coverage is unknown.

api/infrastructure/litellm/client.py
```diff
@@ -39,15 +49,160 @@ def _headers(self, master_key: str) -> dict[str, str]:
...
+        """
+        try:
+            info = self._team_info(org_id, self._headers(self._master_key()))
+        except (httpx.HTTPError, ValueError, KeyError, TypeError) as exc:
+            raise LiteLLMError("Failed to read Organization team spend") from exc
+        if info is None:
+            return None
+        return {"spend": info.get("spend"), "renews_at": info.get("budget_reset_at")}
+
+    def get_key_team(self, key: str) -> str | None:
```
kalibetre, line 151: **Consider — send the key's hash, not the key.** I checked LiteLLM v1.96.2: `/key/info` runs `_hash_token_if_needed`, which passes any value that doesn't start with `sk-` through unchanged, and `hash_token` is `sha256(token).hexdigest()`. Sending the hash would keep the plaintext key out of query strings and access logs entirely. To be fair, dropping the exception chain here is already better than the existing `get_key_info`, which puts `{exc}` in its message.

api/domains/organizations/service.py
```diff
@@ -50,12 +75,361 @@ def _and_list(items: list[str]) -> str:
...
+                # One unreadable Organization must not cost every other one its check.
+                logger.exception("LLM budget threshold check failed for Organization %s", organization.id)
+                continue
+            for crossing in crossings:
+                # Recorded only once the notification is staged. The other order means
+                # a failed publish marks the threshold alerted and it never fires
+                # again for this window — the alert is lost, not delayed.
+                if not self._publish_budget_crossing(crossing):
+                    continue
+                organization = crossing.pop("organization")
```
kalibetre: **Consider — `list[dict]` with the ORM object tucked inside, then `pop`ped.** A small dataclass (`BudgetCrossing`) holding the organization and the payload fields would remove the smuggled entry and the mutation of the returned items.

api/domains/organizations/service.py
```diff
@@ -50,12 +75,361 @@ def _and_list(items: list[str]) -> str:
...
+                    payload={
+                        "organization_id": organization_id,
+                        "threshold_percent": crossing["threshold_percent"],
+                        "spend_usd": crossing["spend_usd"],
+                        "limit_usd": crossing["limit_usd"],
+                        "renews_at": crossing["renews_at"],
+                        "subject_display": crossing["organization_name"],
+                    },
+                )
+                self.organization_repository.outbox_repository.stage(
```
kalibetre: **Consider — duplicated stage-and-select block.** This is the second copy of the block at `staging:api/domains/organizations/service.py:327` (the model-allowlist code, `service.py:709` on this branch). Both put a `select(EventDelivery.id)` in the service, which `docs/guidelines/code.md:31` forbids ("Services MUST NOT embed SQL"). Worth extracting one repository helper used by both. The two recipient finders in `organizations/repository.py` also duplicate the same dedupe loop.

ui/src/features/organizations/hooks/use-organization-llm-budget.ts
```diff
@@ -0,0 +1,35 @@
...
+import { useActiveOrgRole } from "./use-active-org-role";
+
+/** The Organization's own view of its spend limit.
+ *
+ *  Served from a stored snapshot rather than the provider, so it costs a single
+ *  local query on page load. Skipped entirely for members: the endpoint requires
+ *  cost.read, so fetching would only ever produce a 403. */
+export function useOrganizationLlmBudget() {
+  const { canManage, selectedOrganization } = useActiveOrgRole();
+  const organizationId = selectedOrganization?.id ?? null;
```
kalibetre, line 21: **Nit — hand-built query key.** `organizationsKey` exists (`utils.ts:4`) and `webapp.md` says to use centralized query keys.

ui/src/features/costs/components/costs-page.tsx
```diff
@@ -15,6 +15,8 @@ import { AgentsBySpend } from "./agents-by-spend";
 import { CostChartsPanel } from "./cost-charts-panel";
 import { CostFilterBar } from "./cost-filter-bar";
 import { CostList } from "./cost-list";
+import { LlmBudgetBanner } from "@/features/organizations/components/llm-budget-banner";
```
kalibetre: **Nit — import order.** The `@/features/...` alias import sits after the relative imports, with a stray blank line splitting the relative group.

ui/src/features/organizations/components/llm-budget-card.tsx
```diff
@@ -0,0 +1,250 @@
...
+            >
+              Remove limit
+            </button>
+          )}
+        </div>
+      )}
+
+      {invalid && !gated && (
+        <p className="m-0 mt-2 text-[12.5px]" style={{ color: "var(--err)" }}>
+          Enter a positive amount, or clear the field to remove the limit.
```
kalibetre: **Nit — copy says "positive amount", but `0` is valid** (`parsed < 0` is the check, and the service handles `limit == 0` explicitly). Something like "Enter an amount of zero or more" would be accurate.

helm/agentbarn-api/templates/deployment.yaml
```diff
@@ -79,6 +79,8 @@ spec:
               value: "http://{{ .Release.Name }}-communications:{{ .Values.communications.service.port }}/communications/v1"
             - name: AGENT_IMAGE_PULL_SECRET
               value: "{{ .Values.agentImagePullSecret }}"
+            - name: ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS
```
kalibetre, line 82: **Nit — PR description.** It says `deployment.yaml` is "byte-identical to staging again", but this adds the two threshold env lines. Worth correcting the description.

kalibetre (changes requested):
**Request changes.** The design is sound: authorization is tight, "unknown is not zero" is handled, the row stays authoritative when the proxy push fails, and the migrations and docs are thorough. Two things need fixing before merge, and a few more are worth doing.

**Required**
1. Two Alembic heads once merged (inline on `97f14c9b7000_merge_heads.py`).
2. `ORGANIZATION_LLM_BUDGET_ALERT_THRESHOLDS` isn't exported by `deploy.yml` / `deploy-public.yml`, so the GitHub Variable is inert (inline on `helmfile.yaml.gotmpl`).

**Should fix**
3. A threshold list without `100` never fires the exhausted alert.
4. The coverage/enroll sweep repeats a K8s Secret read and proxy call per Agent, and every budget save reruns it.
5. Native `<select>` against `webapp.md:7`.
6. `OrganizationService` grew from 368 to 750 lines. Budget, enrollment, coverage, alerting and reconciliation could move into a separate budget service; the two 26-line cron entry modules (`llm_budget_alerts.py`, `llm_budget_reconciliation.py`) are near-clones.

**Consider / nits**: see the inline comments. Also: `CONTEXT.md` has no term for any of this, and the UI, emails and docs use "limit", "allowance" and "budget" interchangeably. The PR is 53 files / +4107 and includes #191's commits, so the in-pod error rewrite and the UI banners could split out.

**Verification**: I ran the two new unit test files locally (102 passed) and checked LiteLLM behaviour against the v1.96.2 source (`/team/update` keeps explicit nulls; `/key/info` accepts a hashed key). I did not run the integration tests, Playwright, Helm or lint, or anything against a live LiteLLM.

kalibetre (approved):

PR #207, opened 2026-09-16
AF-298: add restore point settings and atomic configuration replay

ddlaurin:
Ticket 3 of the AF-182 epic. Depends on AF-292, which is merged.

## What this adds

A "Restore points" section on the Agent configuration page, between "Agent-owned
override" and "Danger zone". From there you can capture a restore point, restore one,
or delete one.

Each entry shows the configuration the Agent had when the archive was made, next to
its configuration now. When you restore, you can also choose to put that old
configuration back.

Restore replaces the Agent's files, so the dialog says so clearly. You have to type
the Agent's name, you get a list of what will be replaced, and the dialog explains
that the runtime's own chat history goes back with the files while the conversation
history stored here does not.

## API changes

Three things the ticket expected to be ready were not.

**1. The manifest was almost empty.**

Ticket 1 says it records the template version and the skill versions. It recorded
neither, so there was nothing to compare and nothing to put back. It now records
both, and the manifest version is 2.

It also records where the template comes from: the platform, the organization, or the
Agent's own override. It has to, because an organization copy of a platform template
keeps the same key and starts again at version 1. Without the source, two different
templates look identical.

**2. The UI could not find out the limit.**

An Agent can only keep a set number of restore points. That number ignores automatic
backups and failed captures, but the list `total` counts them both. So the UI had no
way to work out how many were left. The list now returns `cap` and `manual_count`.

**3. Putting the old configuration back could not work as two requests.**

`select_agent_template` checks the template's required skills against the skills the
Agent has right now. `update_agent` checks the skills against the template the Agent
is pinned to right now. Whichever one you send first is refused, because the other
half has not been saved yet.

So an old template that needs a skill at version 1 was refused whenever the Agent had
moved that skill to version 2, even though nothing was missing.

`select_agent_template` now takes the skill versions and the settings that belong with
the template. It checks them against the skills the Agent will have after the change,
and saves the template, the skills and the settings in one database transaction. The
whole configuration is applied, or none of it is.

## One change to read carefully

`_validate_override_requirements` is also used when publishing an override and when
saving a draft. It used to require every skill in the template's list to be assigned
at the exact version. That made "one of these skills" groups impossible: the skill you
did not choose failed the version check, and the code even asked for its credentials.

It now follows the group rule, and it uses `_validate_required_skill_versions` for
versions, which is the check `update_agent` already uses. Both paths now accept the
same configurations.

## Notes for deployment

- No database migration, no chart change, no new environment variables.
- The manifest is stored in a JSON column and read back as it is. That makes a
  rollback safe both ways: older code returns version 2 rows untouched, and this code
  shows a version 1 row's missing fields as "Not recorded", which is not the same as
  empty.
- `agent.updated` is now also sent when a template selection carries settings. Today
  only the configuration replay does that. It creates a Security Audit Record.
- The API and the UI are separate charts in one release. While they roll out, a new UI
  pod talking to an old API pod shows the section's error state, because the UI needs
  `cap`. It clears once the API finishes rolling out.

## Testing

API tests cover: a template and its skills applied together (the case two requests
cannot handle), choosing the right template when a platform template and an
organization copy share a key and version, a group where one skill sits at a different
version, a refused request leaving everything unchanged, the Agent's timestamp moving
when only skills change, and `null` settings refused as a validation error instead of
reaching the database.

UI: 17 Playwright tests. Capture, waiting for it to become ready, restore with the
typed name, the limit being reached, loading more than one page, hiding the replay
option without `agent.update`, showing a change the label alone cannot show, and the
dialog staying open while the replay is still running.

The full API and UI suites and the lint checks pass in CI.

## Docs

`docs/features/agents.md` now describes the section, the manifest and its versions, how
the template source is stored and checked, and that the configuration is applied in one
go. The epic log is updated.


ui/src/features/agents/components/restore-point-restore-dialog.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+    try {
+      await restore.mutateAsync(restorePoint.id);
+    } catch (error) {
+      toastError(error);
+      return;
+    }
+    setRestoreAccepted(true);
+
+    if (reapply && replayable) {
```
AnaniyaT: **Restore and re-apply should be a single request handled by the server**

Right now the dialog sends the restore and the configuration as two separate requests, and nothing checks beforehand whether the configuration can still be applied. This causes two problems:
- **If the template, a skill version or the model is no longer available:** the configuration request fails, but only after the files are already being overwritten.
- **If the restore Job fails:** the configuration has already been changed anyway. The automatic backup holding the old configuration is marked FAILED, so there's no way back from the UI.

Could `POST …/restore` take `reapply_configuration` instead? The server would check the configuration before starting the Job, refusing if it can't be applied. It would apply the configuration through `select_pin` only after the files are restored. It's the same validation and write path the ticket asks for.

api/domains/agents/service.py
```diff
@@ -1261,19 +1291,52 @@ def select_agent_template(
...
+                removed_skill_ids=data.removed_skill_ids,
+                scalar_updates=scalar_updates,
             )
         except AgentOverrideConcurrencyError as exc:
             raise HTTPException(status_code=status.HTTP_409_CONFLICT, detail=str(exc)) from exc
         except ValueError as exc:
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=str(exc)) from exc
         return self._get_agent_read(selected_agent, context)
 
+    def _validate_prospective_skill_providers(
```
AnaniyaT: **Credential check blocks unrelated template switches**

This runs on every template selection and checks every skill the Agent has. A skill's required credentials are updated when a new version is published (`skills/repository.py:615`). An Agent on an older version can therefore be blocked from switching templates by a skill it didn't touch. Can we limit the check to the skills added or re-pinned in this request, like `update_agent` does?

ui/src/features/agents/components/restore-point-restore-dialog.tsx
```diff
@@ -0,0 +1,282 @@
...
+          // The recorded scope, not a lookup by key: a fork shares its platform
+          // lineage's key and restarts at v1.
+          selectionType: manifest.templateSelectionType as "platform" | "organization",
+          templateKey: manifest.templateKey,
+          templateVersion: manifest.templateVersion,
+        });
+      }
+      return null;
+    } catch (error) {
+      return error instanceof Error
```
AnaniyaT: **Replay errors show IDs**

Errors reach the user with UUIDs (`Skill <uuid> not found`, `Version 3 not found for skill <uuid>`). The manifest has the skill names, so can we show those instead?

ui/src/features/agents/components/agent-restore-points-settings.tsx
```diff
@@ -0,0 +1,356 @@
...
+                style={{ border: "1px solid var(--line)", background: "var(--bg-elev)" }}
+                data-testid="restore-point-row"
+              >
+                <div className="flex flex-wrap items-start justify-between gap-3">
+                  <div className="min-w-0">
+                    <div className="flex flex-wrap items-center gap-2">
+                      <span className="text-[0.88rem] font-medium" style={{ color: "var(--ink)" }}>
+                        {restorePointLabel(restorePoint)}
+                      </span>
+                      {isSystem && (
```
AnaniyaT: **Badge**

Minor: every automatic backup gets the same "System-created" badge. "Before restore" would say what it's for.

ddlaurin:
@AnaniyaT 
1. Done as you suggested. POST .../restore now takes reapply_configuration. The server checks the recorded configuration before starting the Job and refuses if it can't be applied, so no files are touched. The configuration is written through select_pin only after the Job confirms the volume is back, so a failed Job never changes it. The browser no longer writes anything. The request is stored on the restore point and cleared in the same transaction as the write, so it isn't lost if the process stops, and it runs under the Agent's lifecycle lock. The selection is rebuilt on the server from the stored manifest, so what is checked before the Job is what is written after it. Two side effects: while a replay is still waiting the Agent can't be started, and it gets applied on the next read of the list. If it can't be applied by then, the restore still stands and the row says why, with a button to try again. There's a new migration for the stored request.
2. The check now only covers skills a request adds or re-pins, like update_agent. There's a test for your exact case.
3. Skill IDs in the message are replaced with the names from the manifest.
4. Each kind has its own label now: "Before restore", "Before reset", "Before upgrade". Manual captures have no badge.

AnaniyaT (dismissed):

PR #208, opened 2026-09-16
AF-329 Surface provider credit failures across communication channels

kalibetre:
## Summary
- Detect and name exhausted provider credits in user-facing failure notices.
- Post actionable provider failure reasons in web chat and threaded Slack replies.
- Extend consistent failure feedback across Slack, Discord, Microsoft Teams, and Telegram.


api/domains/communications/plugins/discord.py
```diff
@@ -199,6 +203,42 @@ def send(
...
+        if context.stage != ProcessingFeedbackStage.FAILED:
+            return
+        try:
+            DiscordClient(credentials.bot_token).send_message(
+                context.location.id,
+                failure_notice(context.error_summary),
+                reply_to_id=context.provider_message_id,
+                # One notice per dead-lettered Delivery, even if the hook re-runs.
+                idempotency_key=(
+                    provider_idempotency_key(str(context.source_delivery_id))
```
ddlaurin: The reply already uses the same idempotency key: in `communications-runtime-adapter.py:189`,
`post_reply()` sets `idempotency_key = delivery_id` when there is no suffix.
Both go through `provider_idempotency_key()`, so both messages get the same
Discord nonce.

We also send `enforce_nonce: True`. With this flag Discord does not create a second message.

Scenario:

1. The agent writes a reply. Sending it to Discord fails, and all retries fail. The delivery becomes DEAD_LETTERED
2. The new code posts the failure notice. Its nonce is `hash(<source delivery id>)`.
3. A user opens the connection diagnostics and clicks "Retry" on that delivery.
4. The retry sends the reply again, with the same nonce.
5. Discord returns the notice message and does not post the reply.
6. We mark the delivery as SUCCEEDED.

The user never sees the agent's answer, but we would record the delivery as delivered.

ddlaurin (approved):

PR #210, opened 2026-09-16
AF-320: trigger an agent from an external system over HTTP

sami-bre:
Closes the walking skeleton of [AF-304](https://aai-labs.atlassian.net/browse/AF-304). A user adds a webhook connection to an agent and gets a URL and a generated signing secret. Any external system — Jira automation, a CI job, another agent — can then POST a signed request whose `prompt` field *is* the instruction, and the agent runs it with no human in the loop.

Ticket: [AF-320](https://aai-labs.atlassian.net/browse/AF-320).

**This PR revises the original implementation** (see commit history) after review. The linked architecture doc and the original PR description below described the connection-level prompt template, bearer auth, and a user-supplied secret — all three are now gone. The doc has not been updated yet; treat this PR body as the current source of truth.

## What changed in the revision

- **The caller supplies the instruction directly.** The webhook body is `{event_id, prompt, ordering_key?}` — no connection-level `{{ payload.x.y }}` template, no separate `payload` object. `prompt` becomes the envelope's `text` directly, so the transcript shows exactly what was asked. This does widen the trust model: whoever holds the secret can now ask the agent to do anything it's capable of, not just the one job a fixed template would allow. Accepted trade for the simplicity; replay is still bounded by `event_id` idempotency.
- **HMAC-SHA256 only.** The bearer-token option is removed; every real caller can sign a body.
- **The signing secret is minted by the server and shown exactly once**, via new `PlatformPlugin.mint_credentials()` / `reveal_once()` hooks (generic seam, webhook is the only platform that uses them today). A new `POST .../connections/{id}/rotate-credentials` endpoint re-mints and reveals a new secret without changing the URL — the answer to "I lost the secret."
- **An agent can hold more than one webhook connection** — one per calling system. Every other platform keeps today's one-active-connection-per-platform rule: a new nullable `singleton_key` column (a copy of `platform_key`, `NULL` only for platforms that opt in via `allows_multiple_connections`) replaces the old platform-keyed unique index, so nothing about Slack/Discord/Teams/Telegram/email changes.
- **Webhooks get their own "Webhooks" configuration tab** (list → detail), out of Messaging — a machine trigger isn't a way to message the agent, and the "nobody can message this agent" nudge no longer counts a webhook as messaging.
- **Webhook calls no longer show up in the Conversations tab.** A request/response pair with a status and an error reads as a log line, not a conversation. Each webhook's detail view instead lists its calls (prompt sent, response received, error if any), backed by a new generic `GET .../connections/{id}/calls` endpoint that pairs each inbound delivery with the agent's reply via `OutboundCommunicationEnvelope.source_delivery_id`. The underlying `agent_chat_message` row is still written (the delivery's `message_id` FK is `NOT NULL`, and platform activity stats read that table directly) — it's just not surfaced in that view anymore.

## Carried over from the original implementation, unchanged

- `kind` and `session_key` on `communication_delivery`, `EVENT` on `ConversationType`.
- Protocol version 3, so a pod running the previous adapter never gets handed an `EVENT` (stays `PENDING` until restart) instead of mishandling it.
- Events run in a fresh session, approvals off, progress suppressed; a delivery the agent couldn't start is *released*, not completed.
- `execution_policy.py` stays the only place a `DeliveryKind` is ever compared — everywhere else reads a field off a policy.
- `CONVERSATION` behavior is untouched throughout; `test_communication_deliveries.py` is the regression gate and passes byte-for-byte.

## Verified against a live stack

Ran end to end on local k3d against a real agent: created two webhook connections, copied the one-time secrets, fired signed requests (HMAC verified, wrong-secret rejected, missing-field 400s), watched the calls show up in each webhook's own calls list with the right prompt/error/response, regenerated a secret and confirmed the old signature stopped working while the URL stayed the same, and confirmed the Conversations tab and Messaging tab both correctly show no trace of the webhook connections or their calls.

## Gates

| | |
|---|---|
| `make check-api` | pass |
| `make check-ui` | pass |
| `make check-migrations` | pass (single head) |
| Migration up → down → up | verified against a disposable Postgres |
| Targeted `pytest` (webhook plugin, webhook triggers, connections, conversations, gateway, email plugin) | all passing |
| `make test-api` / `make test-ui` (full suites) | not run locally this round — leaving to CI |

Migration `43eefac934c1` adds `singleton_key`, backfills it from `platform_key` for every row except `webhook`, and swaps the unique index. Downgrade reverses it, reusing the original migration's duplicate-row guard.

## Out of scope

Replies to a declared `response_url`, status lookup and the test button (AF-322); queue-when-stopped and the retry split (AF-321); rate limits and budgets (AF-323); per-delivery approvals (AF-324). Still no concurrency cap on parallel event runs per pod — unchanged from the original PR, now more reachable with multiple webhooks per agent.

[AF-304]: https://aai-labs.atlassian.net/browse/AF-304?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AF-320]: https://aai-labs.atlassian.net/browse/AF-320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dominykas-aai-labs:
### Must fix

**1. An unauthenticated caller can get internal error text back, including stored credentials** (`gateway_routes.py`, `accept_provider_webhook`)
- The new `except ValueError` returns `str(exc)` as a 400. Pydantic's `ValidationError` is a subclass of `ValueError`.
- `gateway_service.accept_provider_webhook` validates settings and credentials *before* checking the signature. If a stored credential no longer fits its model, an unsigned POST gets pydantic's message back, and that message includes `input_value`, which is the decrypted secret.
- It also turns any `ValueError` on the Teams path from a 500 into a 400.
- **Fix:** add a dedicated exception (e.g. `WebhookRequestRejected`) in `plugins/base.py`. The webhook plugin raises it for version and contract errors, and the route catches only that.

**2. PATCH still lets a user choose their own signing secret** (`service.py`, `update_connection`)
- `data.credentials` is validated and stored as given, so any string of 32+ characters becomes the secret. The ticket says the secret is server-generated only.
- **Fix:** if `plugin.mint_credentials()` is non-empty and `data.credentials` is set, return 400 and point to `rotate-credentials`. Add a test for it.

### Should fix before merge

**3. The migration and the model disagree about an index** (`d5e7a1c3b902_delivery_kind_and_session_key.py`)
- `ix_communication_delivery_kind_status` is created in the migration but not declared in `CommunicationDelivery.__table_args__`, so the next autogenerate will emit a drop for it.
- `make check-migrations` only counts heads, so nothing catches this.
- The index also adds little: `kind` has two values and the claim query already narrows by agent, direction and status. It is built without `CONCURRENTLY` on a hot table.
- **Fix:** remove it from the migration; the branch isn't merged, so editing the migration is safe.

**4. Nothing tests the Teams webhook through the new raw-body route**
- The acceptance criteria call this out: "including the Teams webhook path after the raw-body change". No test posts a Teams activity to `/webhooks/{connection_id}`.
- Behaviour also changed: a missing `Authorization` header used to give a 422 and now gives a 401.
- **Fix:** add an integration test with `verify_inbound_jwt` patched. A valid activity should return 202 with an accepted delivery; a missing or bad Authorization header should return 401.

### Lower priority

**5. `release_runtime_delivery` claims a bound it doesn't have**
- A claim adds 1 to the attempt count and a release subtracts 1. A delivery that only ever hits "busy" therefore never reaches `max_attempts`, although the docstring says it is bounded.
- A cancel request is recorded as `RELEASE_LIMIT` instead of going through the cancel outcome.
- In practice an EVENT only hits "busy" when the same delivery is reclaimed while its first run is still going, so the loop ends when that run finishes.
- **Fix:** send cancellation through the existing cancel path and drop the "bounded" claim.

**6. Leftovers from the template and bearer revisions**
- Both implementations of `verify_webhook(settings, ...)` discard `settings`, and the gateway now validates settings before auth only to pass them in. Revert to `verify_webhook(credentials, request)`.
- `agent-channel-settings.tsx` has `format: "textarea"` support that no plugin uses any more.
- Several comments and names still refer to `subject`, which was dropped:
  - `_message_values` ("groups the timeline by subject")
  - the `list_calls` docstring ("per-subject")
  - the test `test_events_sharing_an_ordering_key_serialise_even_across_subjects`
- The webhook module docstring still says "a named job".

**7. The body-size cap is checked after the whole body is read**
- `await request.body()` reads everything before the 256 KiB check.
- Body limits are out of scope for this ticket, but the code presents this as a protection.
- **Fix:** reject early on a `Content-Length` over the cap and keep the post-read check.

**8. Nit: comment volume**
- The diff adds a lot of rationale comments and multi-paragraph docstrings: `execution_policy`, `webhook.py`, the adapter, `delivery_repository`, the models, the migrations and the TSX files.
- It also adds an unrelated DM comment in `conversations-tab.tsx`.
- Suggest trimming these and moving the rationale into the PR description.

dominykas-aai-labs (approved):

PR #218, opened 2026-09-17
AF-302: connect SharePoint by signing in on the agent's Teams app

AnaniyaT:
Give agents SharePoint access through a Microsoft sign-in on the agent's existing Microsoft Teams app, stored like any other credential and refreshed by aai-cli's own `microsoft_delegated` profile — no app of ours, no publisher verification, nothing Agent Barn–specific in aai-cli.

## What's included

- **Sign-in (`api/domains/agents/sharepoint_service.py`, `microsoft_identity.py`, `integrations/microsoft_oauth/routes.py`)** — popup sign-in on the Teams app as a public client (PKCE, never the app's secret). New `setup`, `authorize-url` and `sign-in` routes under the agent, plus the fixed `/integrations/microsoft/callback`. The sign-in checks the signed state, the user who started it, the tenant (GUID in any case, or a tenant given as a domain), the SharePoint permission and a refresh token, then saves the `sharepoint` credential. Microsoft errors map to fixes: administrator approval needed, declined consent, redirect URI under Web / public client flows off.
- **Communications (`communications/service.py`)** — `get_teams_app_identity` exposes only the Teams app's id and tenant; the app secret stays in the domain.
- **Credential (`agents/models.py`, migration `e2b7c4d19a58`)** — `sharepoint` provider (connection, tenant, client id, account, granted permissions, access level, refresh token, per-sign-in id). The generic secret upsert rejects it; shared credentials don't allow it.
- **Runtime (`aai_cli_artifacts.py`, `agents/service.py`)** — `[profiles.sharepoint-work]` with `provider = "microsoft"`, `auth_type = "microsoft_delegated"`, scope limited to SharePoint. The boot script writes the refresh token only for a new sign-in (so restarts keep the token aai-cli rotates) and removes it once SharePoint is removed. OpenClaw's aai-cli store moves onto its volume (`/home/node/.openclaw/aai-cli`), like Hermes'; other providers are unaffected since their secrets are rewritten every boot. The boot script is now mounted for every agent, installing config only when there are profiles.
- **Skill** — bundles aai-cli's `aai-microsoft` skill (shown as SharePoint); the integrations text tells the agent it has SharePoint only.
- **UI (`sharepoint-sign-in.tsx`, `use-sharepoint-sign-in.ts`, `sharepoint-setup.ts`)** — sign-in panel in the skills tab and the Integrations section with a guided checklist: direct Microsoft Entra links for the redirect URI (Mobile and desktop applications), public client flows and the Graph permission, Microsoft's guide for each step, and an administrator approval link to copy or use in place. The hire dialog explains a template requiring SharePoint can't be hired yet. Google sign-in now shares the popup handling (`use-oauth-popup.ts`); its behaviour is unchanged.
- **Docs** — `docs/features/integrations.md` SharePoint section.

## Required repo config

None.

## Notes

- Customer setup on the Teams app (all shown in the UI): redirect URI `{web_app_url}/api/v1/integrations/microsoft/callback` under **Mobile and desktop applications**, **Allow public client flows: Yes**, Microsoft Graph delegated `Sites.ReadWrite.All` (or `Sites.Read.All`), and admin consent where the organization restricts user consent (aai-labs does).
- The agent reaches every site the signed-in account can open; there's no per-site limit.
- Microsoft refresh tokens expire after 90 days unused, so an agent that doesn't touch SharePoint for that long needs a reconnect.
- The staging base images were rebuilt after aai-cli's Microsoft Graph integration merged, so no base image change is needed here.
- Out of scope: agents without Teams, template-required SharePoint, shared credentials, other Microsoft 365 services.

## Testing

- `make check-api`, `make check-migrations`, `make check-ui`, `make lint-ui`
- `make test-api` — all pass except the known `test_email_connection_is_refused_when_the_environment_has_no_agent_email_domain`, which fails only from the local `.env` and passes with `AGENT_EMAIL_DOMAIN` emptied
- `make test-api-k8s` (against local k3d), `make test-api-runtime` (with the locally built Hermes image), `make test-ui` (294 passed)
- Migration upgrade → downgrade → upgrade on a fresh database
- Live on local k3d with the aai-labs Teams bot app: sign-in, then read and create files on `aailabs.sharepoint.com/sites/merkys-test-1` from a **Hermes** and an **OpenClaw** agent; both still authenticated after a pod restart with the rotated token kept
- CI green on the branch


api/domains/agents/service.py
```diff
@@ -2096,14 +2098,25 @@ def _start_agent_unchecked(self, agent: Agent, actor: ActorIdentity) -> Agent:
...
             )
-        store = {p: c for p, c in decrypted.items() if p.value in provider_secrets_map}
+        # SharePoint's refresh token goes through the store too, written only for a new sign-in.
+        store = {
+            p: c for p, c in decrypted.items() if p.value in provider_secrets_map or p == SecretProvider.SHAREPOINT
+        }
         aai_home = "/opt/data" if agent.agent_type == AgentType.HERMES else "/home/node"
+        # The store must survive restarts: aai-cli rotates delegated Microsoft tokens in it.
+        # Hermes' home is its volume; OpenClaw's volume is only ~/.openclaw.
+        aai_store_dir = None if agent.agent_type == AgentType.HERMES else "/home/node/.openclaw/aai-cli"
```
sami-bre, line 2108: **Blocking: restore points would now include the aai-cli secret store and its key.**

This puts OpenClaw's aai-cli store on the agent's volume for every provider, not just SharePoint. But `OPENCLAW_EXCLUDED` in `api/domains/agents/restore_point_job.py` has no `aai-cli` entry. Hermes already excludes `.config/aai-cli`.

So a restore point of any OpenClaw agent with an aai-cli integration (GitHub, Jira, Slack, SharePoint…) will hold `aai-cli/aai-secrets.enc.json` and `aai-cli/key` side by side: every token plus the key to read them. `docs/features/agents.md` says the archive excludes credential material.

Suggested fix: add `"aai-cli"` to `OPENCLAW_EXCLUDED`, plus an OpenClaw version of `test_hermes_capture_excludes_the_aai_cli_credential_store`.

api/migrations/versions/e2b7c4d19a58_add_sharepoint_to_secret_provider.py
```diff
@@ -0,0 +1,41 @@
...
+_OLD is copied verbatim from c9f1b30a7d42 (remove_legacy_google_providers), the most recent
+revision to recreate the provider constraint — not reconstructed from an older one.
+5e0adff0f5e2 exists because two branches each rebuilt this constraint from a stale base and
+silently dropped a provider that was already in use.
+"""
+
+from alembic import op
+
+revision: str = "e2b7c4d19a58"
+down_revision: str | None = "c8e1f4a9b2d6"
```
sami-bre, line 15: **Blocking once rebased: this adds a second Alembic head.**

`staging` has moved past `c8e1f4a9b2d6`; its head is now `4b4bb4af4d31`. After a rebase, `alembic heads` shows both `4b4bb4af4d31` and `e2b7c4d19a58`, and `make check-migrations` fails in CI.

Please set `down_revision` to `4b4bb4af4d31` (or whatever `staging`'s head is when you rebase). The provider list here is still right: nothing on `staging` has changed `ck_agent_secret_provider`.

api/domains/agents/sharepoint_service.py
```diff
@@ -0,0 +1,387 @@
...
+    (``contoso.onmicrosoft.com``); Microsoft reports ``tid`` as a lower-case GUID. A domain
+    can't be compared, but the sign-in already used that tenant's authority, so any tid
+    issued there is that organization.
+    """
+    if not token_tid:
+        return False
+    try:
+        return UUID(configured) == UUID(token_tid)
+    except ValueError:
+        return True
```
sami-bre, line 102: **Non-blocking hardening:** this branch accepts *any* tenant that isn't a GUID, not only a domain as the docstring says. `TeamsCredentials.tenant_id` has no format check (`min_length=1` only). So if a value like `organizations` ever got stored (or anything at all, with `skip_teams_token_validation` on), the tenant check here would pass for an account from any org. This is the check behind "an agent can't reach SharePoint outside its own org".

It isn't reachable with a normal single-tenant Teams app, but it's cheap to close: only treat the value as a domain when it looks like one (for example, it contains a `.`), and return `False` for other non-GUID values.

ui/src/features/agents/components/sharepoint-sign-in.tsx
```diff
@@ -0,0 +1,329 @@
...
+        namePrefix={`sharepoint-${agentId}-`}
+        onFieldChange={onFieldChange}
+        onListChange={() => undefined}
+        disabled={disabled || isSigningIn}
+      />
+
+      <Note>
+        Signing in saves SharePoint access for this agent straight away, even if you don&apos;t apply
+        your other changes. To take it away later, remove SharePoint from the agent&apos;s Integrations.
+      </Note>
```
sami-bre, line 305: **Non-blocking:** if someone signs in and then clicks Cancel instead of Apply, the credential is saved and shows as configured, but a running agent won't use it until its next restart. Could the note say that? For example: "…It takes effect when the agent restarts; Apply does that for you."

sami-bre (commented):
Two blocking issues and two small non-blocking notes, inline.

The blocking ones: restore points would now copy OpenClaw's aai-cli secret store and its key (for every OpenClaw agent, not just SharePoint ones), and the migration will make a second Alembic head once this is rebased on `staging`.

sami-bre (approved):
LGTM now

PR #219, opened 2026-09-17
fix(agents): Harden Agent runtime persistence, scheduled delivery, and OpenClaw startup recovery

kalibetre:
**Summary:**

Several runtime reliability gaps were addressed across Hermes and OpenClaw.

OpenClaw Agents could crash-loop after a runtime upgrade when their PVC contained legacy workspace state. Its implicit 30-minute heartbeat could also generate model usage outside explicit schedules, while the Code Reviewer fallback turned unnamed heartbeats into PR scans. Restore points could retain npm symlinks into the runtime image and then fail validation.

Hermes needed an explicit safe-write contract for both persistent mounts, accurate boot context for persistent memory paths, expected handling for cron sessions without an originating chat, and setup guidance for scheduled Slack delivery.

Scope
- Migrate legacy OpenClaw workspace state only when runtime-owned markers exist.
- Disable ambient OpenClaw heartbeats and replace PVC-held heartbeat configuration.
- Restrict Code Reviewer PR scans to the explicit review-health-scan cron.
- Exclude regenerated OpenClaw npm state and symlinks from restore-point archives.
- Permit Hermes writes only to the intended /opt/data and /workspace mounts.
- Document Hermes persistent memory and daily-note paths in startup context.
- Treat Hermes boot and cron sessions as expected no-chat telemetry cases.
- Document the Slack default delivery target requirement for scheduled Code Reviewer output.


dominykas-aai-labs:
### Should fix

**1. OpenClaw: `doctor --fix` runs before the config and plugin directories are prepared** (`api/domains/agents/scripts/openclaw/start.sh`)
The migration block runs `openclaw doctor --fix --non-interactive` against the stale PVC `openclaw.json`. It runs before `init-openclaw.js` and before the `local-plugins/*` copies. The same script notes that "every `plugins.load.paths` entry must exist before any openclaw CLI call validates the config". Doctor may also rewrite the stale config, which init then deep-merges. Because of `set -e`, a non-zero doctor exit brings back the crash loop this PR is fixing.
→ Move the migration block after the plugin-directory copies and before `OPENCLAW_VERSION=$(openclaw --version …)`.

**2. Nothing shows that a legacy PVC actually starts**
`test_startup_migrates_only_known_legacy_workspace_state` only checks that strings appear in `START_SH`. It proves neither "only when markers exist" nor "starts successfully". The openclaw-base smoke check only greps `doctor --help` for the flags.
→ Add an openclaw-base smoke check that:
- seeds a legacy marker;
- runs the detect-and-migrate step;
- asserts doctor exits 0 and the markers are gone.

Add a negative case too: on a clean workspace, doctor must not run.

**3. The Hermes `cron_` session-ID prefix is assumed, not verified** (`api/domains/agents/scripts/hermes/plugins/telemetry-push/__init__.py`)
The unit test makes up `cron_bb2d5e0d12de_20260917_090007`, and nothing in the repo ties that format to Hermes' scheduler. If upstream changes the format, the warning comes back and no test catches it.
→ Add a hermes-base smoke assertion against `cron.scheduler`, which the smoke test already imports for other cron contracts.

**4. The OpenClaw heartbeat-disable values aren't validated against the runtime**
Nothing checks that OpenClaw treats `{"every": "0m", "target": "none"}` as disabled, rather than rejecting it or falling back to the default. The unit tests only check the builder dict and the literal `REPLACE_PATHS` string.
→ Add a smoke check in openclaw-base that validates a config containing this block and confirms heartbeat resolves to disabled.

### Worth addressing

**5. The Hermes write-root policy is defined twice, and the doc overstates it**
The Dockerfile sets `ENV HERMES_WRITE_SAFE_ROOT=/opt/data:/workspace`. `start.sh` then *merges* those roots into any existing value, so the effective list is "these plus whatever was already set". `docs/architecture/runtime-and-deployment.md` says the roots are set "to allow exactly those roots".
→ Keep one source of truth (the Dockerfile ENV looks sufficient). Otherwise make `start.sh` assign the value outright, then correct the doc.

**6. "File-safety still applies elsewhere" is untested**
`workspace-write-safety-contract` only asserts paths that should be allowed.
→ Add assertions that paths should be denied (e.g. `/etc/...`, `/app/config/...`). Also confirm whether writes to `/tmp` still work now that the allowlist is set explicitly.

**7. The restore-point symlink filter also applies to Hermes** (`api/domains/agents/restore_point_job.py`)
`_walk_included_files` now drops every symlink for both runtimes. That includes relative, in-volume links that `data_filter` would have accepted. The CHANGELOG entry and the test only mention OpenClaw.
→ Either note in the CHANGELOG that symlinks are dropped for every runtime and add a Hermes test case, or limit the filter to OpenClaw.

**8. Test hygiene**
- `assert "HERMES_WRITE_SAFE_ROOT" in HERMES_START_SH` was appended to the unrelated `test_native_gateway_does_not_drain_agent_barn_scheduled_completions`. It should be its own test.
- `test_init_script_replaces_pvc_held_heartbeat_policy` checks for a source string. Running `init-openclaw.js` against a temporary HOME whose config has `heartbeat.every: "30m"` would test the actual replacement.

### Nits

- New rationale comments were added in `openclaw.py`, `restore_point_job.py`, `init-openclaw.js`, both `start.sh` files, the Hermes `Dockerfile` and `smoke-test.sh`. Our convention is to keep that reasoning in commit messages or CHANGELOGs, so please move them there.
- a02077a0 (a Hermes commit) adds the OpenClaw paragraph to `runtime-and-deployment.md`. It belongs with df19d4d6.
- `code-reviewer/agents.md`: "Startup created cron jobs" should be "Startup-created cron jobs".
- 29690416 (Bitbucket skill doc) is outside the ticket's scope. It's fine to keep, but mention it in the PR description.

dominykas-aai-labs (changes requested):
comments are on the PR

kalibetre:
Thanks, all addressed:

1. **Doctor ordering:** moved after the plugin copies and before `OPENCLAW_VERSION=`. A doctor failure is now logged, not fatal.
2. **Legacy PVC test:** new `openclaw-base/test-startup.sh` runs in CI. A seeded legacy PVC migrates (doctor exits 0, markers gone), and a clean PVC never runs doctor.
3. **`cron_` prefix:** the hermes-base smoke test now asserts `run_job` builds `cron_…` session IDs.
4. **Heartbeat:** the same test starts from a stale `30m` config, runs the real init, validates the config, then runs doctor. It asserts the heartbeat job is disabled.
5. **Write roots:** the Dockerfile `ENV` is now the only definition, the `start.sh` merge is removed, and the doc is corrected. I bumped `hermes-base` to `0.2.4`, so Agents get this once they run that image.
6. **Denied paths:** the smoke test now asserts `/etc/passwd`, `/etc/cron.d/…` and `/app/config/…` stay denied. `/tmp` was already denied by the base image's `/opt/data` root, so no regression.
7. **Symlinks:** the filter is limited to OpenClaw. Hermes is unchanged, with a new test and a CHANGELOG note.
8. **Tests:** removed the stray assertion. The heartbeat replacement is now tested by running the real init in the image test.

**Nits:**
- Comments are removed and the rationale is in the commit messages.
- The typo is fixed.
- I'm not rewriting history for the doc-paragraph placement.
- The Bitbucket doc is noted in the PR description.

dominykas-aai-labs (approved):

PR #220, opened 2026-09-20
Af 297 reclaim orphan resources

dominykas-aai-labs:


api/domains/restore_points/reconciliation.py
```diff
@@ -0,0 +1,284 @@
...
+        failed = result.failed
+        for kind, item in candidates:
+            if self._expired(started) or deleted >= RESTORE_POINT_ORPHAN_DELETE_LIMIT:
+                break
+            name = _object_name(item)
+            try:
+                restore_point_id = _labelled_id(item)
+                if restore_point_id is None:
+                    unidentified += 1
+                    logger.warning(
```
AnaniyaT, line 248: **Resources created before this ships can never be reclaimed**

Reclaim keys off the new `agentbarn.io/restore-point-id` label, so any volume or Job created by the code already on staging is skipped — the exact leak this ticket targets. It also warns once per object per run, every 10 minutes, with nothing able to clear it.

Probably a non-issue if nothing has been captured outside local dev yet. Could you confirm? `kubectl -n <ns> get pvc,job -l agentbarn.io/component=restore-point -L agentbarn.io/restore-point-id` — anything with an empty last column is invisible to the reconciler. If there are some, deriving the id from the name (`restore-point-<uuid>`) would pick them up.

api/domains/restore_points/reconciliation.py
```diff
@@ -0,0 +1,284 @@
...
+            return RestorePointReconciliationResult(failed=1)
+
+        resolved = 0
+        replays_attempted = 0
+        failed = 0
+        for row in rows:
+            if self._expired(started):
+                break
+            try:
+                self.service.reconcile_row(row, respect_grace=False)
```
AnaniyaT: **`resolved` overcounts**

A claimed terminal row that owes a replay returns straight out of `reconcile_row`, but still counts toward `resolved`. The summary then reports work that didn't happen.

api/domains/restore_points/reconciliation.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def delete_job(self, name: str, namespace: str) -> None: ...
+
+
+def _labelled_id(item: Any) -> UUID | None:
+    metadata = getattr(item, "metadata", None)
+    labels = getattr(metadata, "labels", None) or {}
+    try:
+        return UUID(labels[RESTORE_POINT_ID_LABEL_KEY])
+    except KeyError, TypeError, ValueError:
```
AnaniyaT: Nit: `except KeyError, TypeError, ValueError:` is valid on 3.14 (PEP 758) and ruff is happy with it, but parentheses would save readers a double-take — I misread it as a Python 2 leftover before checking against the real interpreter.

helm/agentbarn-api/templates/restore-point-reconciliation-cronjob.yaml
```diff
@@ -0,0 +1,67 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "agentbarn-api.labels" . | nindent 4 }}
+    app.kubernetes.io/component: restore-point-reconciliation
+spec:
+  # Must stay longer than RESTORE_POINT_RECONCILIATION_MAX_RUNTIME_SECONDS
+  # (api/domains/restore_points/constants.py). Under Forbid, a run that outlives its
+  # window does not overlap — it silently costs the next tick instead.
+  schedule: {{ .Values.restorePoints.reconciliation.schedule | quote }}
+  concurrencyPolicy: Forbid
```
AnaniyaT, line 15: Nit: no `startingDeadlineSeconds`. Low impact with `Forbid` and the 120-second runtime cap, but a missed window currently just waits for the next tick with nothing recorded.

AnaniyaT (commented):
Reviewed this by running it against a local k3d cluster, not just reading it. Verified: orphaned Job + PVC deleted (pod went Terminating, so Background propagation works), young orphans left alone under the 30-minute floor, labelled resources with a live row untouched, a READY row whose volume was deleted marked FAILED with a clear reason, a 2-hour-old stranded PENDING row resolved, and — the one I cared about most — a restore owing a configuration replay that nobody polled for: the Job succeeded, the row sat at RESTORING owing the replay, and the reconciler then resolved it and applied the configuration. That closes the gap I raised on #207.

Four comments inline, none of them serious.

AnaniyaT (approved):

PR #221, opened 2026-09-21
prod email sending issue

dominykas-aai-labs:


kalibetre (changes requested):
**Verdict: Request changes.** The diagnosis is right and the fix is well aimed, but it reintroduces the bug class it removes, and the required docs are incomplete. This was a static read of the diff and the surrounding workflows; I didn't run CI or `actionlint`.

**Context.** Production inbound email 401s every message. `deploy-public.yml` never passed `AGENT_EMAIL_DOMAIN` or `EMAIL_INBOUND_SECRET` to helmfile, so the chart rendered `""`, and the API rejects every message at `api/domains/communications/gateway_service.py:562` (`if not secret or …`). Separately, the production Worker (`INBOUND_URL = api.agentbarn.dev`) was published by `deploy.yml` from `main`, which deploys the k3s cluster, not the Talos one. I traced the env chain (`helmfile.yaml.gotmpl:172-174` → `helm/agentbarn-api/templates/secret.yaml:27-28`) and it matches the PR's account.

## Required

**1. The `worker.yml` guard can no longer catch an unset staging secret** (`worker.yml:69,88`)

```yaml
INBOUND_SECRET: ${{ inputs.environment == 'staging' && secrets.STAGING_EMAIL_INBOUND_SECRET || secrets.PUBLIC_EMAIL_INBOUND_SECRET }}
```

`A && B || C` yields `C` when `B` is empty. If `STAGING_EMAIL_INBOUND_SECRET` is unset and `PUBLIC_EMAIL_INBOUND_SECRET` is set, the guard passes and the staging Worker is published with the production secret. The staging cluster falls back differently (`deploy.yml`: `… || secrets.EMAIL_INBOUND_SECRET`), so the two sides disagree and every staging message 401s silently. The old guard existed to catch exactly this, and its removed comment described the same trap. The new guard evaluates the same expression it is guarding, so it only fails when *both* secrets are empty. That also contradicts the new `operations.md:78` ("the publish fails when the selected secret is empty").

It also silently selects the production secret for any `environment` other than `staging`, including `""` or a typo.

Suggested fix: validate the specific secret for the target environment and reject unknown environments.

```yaml
- name: Require an inbound secret for the target environment
  env:
    ENVIRONMENT: ${{ inputs.environment }}
    STAGING_SECRET: ${{ secrets.STAGING_EMAIL_INBOUND_SECRET }}
    PUBLIC_SECRET: ${{ secrets.PUBLIC_EMAIL_INBOUND_SECRET }}
  run: |
    case "$ENVIRONMENT" in
      staging)    secret="$STAGING_SECRET" ;;
      production) secret="$PUBLIC_SECRET" ;;
      *) echo "::error::Unknown Worker environment '$ENVIRONMENT'"; exit 1 ;;
    esac
    [ -n "$secret" ] || { echo "::error::No inbound secret set for '$ENVIRONMENT'; every inbound email would 401."; exit 1; }
```

The wrangler step's ternary can stay: once the guard passes, its fallback branch is unreachable. Passing `ENVIRONMENT` via `env:` also stops the current `echo "…'${{ inputs.environment }}'…"` from interpolating an input straight into shell.

**2. The public setup tables in `docs/guidelines/operations.md` weren't updated** (~lines 148-170)

This PR makes these mandatory for the next release tag, but they aren't in "Public GitHub variables" / "Public GitHub secrets":
- `PUBLIC_AGENT_EMAIL_DOMAIN` (variable, e.g. `agents.agentbarn.dev`)
- `PUBLIC_EMAIL_INBOUND_SECRET` (secret)
- `CLOUDFLARE_WORKERS_TOKEN`, which the public `deploy-worker` now consumes. It should be added to the "Shared with k3s" line.

The only mention today is the CHANGELOG "Follow-up" bullet. `AGENTS.md` requires the owning contributor-facing doc to change with CI/deploy behaviour, and the setup checklist is what operators actually follow.

**3. The PR has no description**

The body is empty and the title ("prod email sending issue") isn't standalone. This PR has a hard pre-merge/pre-tag dependency (the two GitHub values and the Cloudflare routing rule for `agents.agentbarn.dev`), which belongs in the PR body rather than only in a changelog line. Suggested title: "Publish the production email Worker from deploy-public.yml with its own inbound secret".

## Nit / FYI

- **Nit:** the `worker.yml:4-5` header still says the workflow is called "by deploy.yml (staging/main, deploy: true)". It's now staging-only from `deploy.yml`, plus `deploy-public.yml`.
- **Nit:** `operations.md:78` "all three values must differ" means three environment pairs; "the three environments' pairs must all differ" reads less ambiguously.
- **FYI:** unlike `deploy.yml`, the public `deploy-worker` isn't gated on `workers/**` changes, so it re-runs the dry-run check and republishes on every release tag once the domain var is set. That's idempotent and convenient for secret rotation, so no change requested.
- **FYI:** `vars` is valid in a job-level `if` on a reusable-workflow call, and gating on the domain var rather than the secret is sound since `secrets` isn't available there.

## Verification

The PR says nothing about how this was verified, and CI can't exercise the tag-triggered public path pre-merge. Please confirm after the first release tag that `deploy-worker` ran green and that a test message to `agent+<slug>-<token>@agents.agentbarn.dev` is accepted, and dispatch `deploy.yml` on `staging` once to prove the staging path still publishes.

kalibetre (approved):

